PR #2, opened 2016-07-18
MODCLUSTER-463 converts aliases for a context on -APP message arrival.

csutherl:
I'm porting this change from 1.3.x into master so that it isn't lost.


csutherl:
Here's a link to the PR in case you wanted to look at it https://github.com/modcluster/mod_cluster/pull/182

PR #13, opened 2017-06-15
MODCLUSTER-591 MemManagerFile should not default to server_root + "/l…

rhusar:
…ogs"


Karm:
@rhusar I'm not 100% convinced this is enough to make it better. Take a look into ```create_attach_mem_node``` and on, e.g. ```ap_log_error(APLOG_MARK, APLOG_NOERRNO|APLOG_EMERG, 0, s, "get_mem_node %s failed", node);``` messages.   It would be great if it suggests user using ```MemManagerFile``` to specify an existing directory...    WDYT?

Karm:
test

Karm:
test

Karm:
run tests

jfclere:
actually the default httpd doesn't have cache, so the fix doesn't look like a good idea.

PR #19, opened 2017-10-27
MODCLUSTER-622 segfault in process_info

hokuda:


Karm:
Can one of the admins verify this patch?

rhusar:
This is ok to test.

rhusar:
@Karm Could we fix the account that the CI integration uses? We have https://github.com/modcluster-pull-request for that.

jfclere:
It is still not prefect: The apr_time_now(); I would rather do it before the the lock and have a local variable in the routine and set it in inside the loop.

hokuda:
@jfclere Thanks for the suggestion. If the last one is ok, I will rebase.

PR #28, opened 2020-08-04
Fix for JBCS-984: Prevent adding NODES with the same scheme/hostname/…

jfclere:
…port those will result in the same worker

and mess up the logic when getting a worker and reusing it...


gzaronikas (approved):
LGTM, Approving

PR #30, opened 2020-11-09
Fix for mod_manager.c,opening files in children processes and shared memory creation [MODCLUSTER-726][MODCLUSTER-725]

kimonides:
mod_manager.c had an issue where it would try to open some .lock files in the manager_child_init function, it wouldn't find them and try to create them but it didn't have the permissions since it was a child process.
I just created the .lock files in the manager_init function so all the child processes could inherit the file descriptors.

EDIT**

Also fixed some problems that mod_manager.c has with shared memories.
It would create a shared memory with apr_shm_create, which if httpd didn't shut down properly was never clean up, as a result when httpd would be started again it would try to apr_shm_create the same file and get file exists error.
Now it tries to create an anonymous shared memory, if the system doesn't support anonymous shared memories it removes the old named shared memory if it exists and creates a new one.

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -570,6 +570,7 @@ static int manager_init(apr_pool_t *p, apr_pool_t *plog,
     char *sessionid;
     char *domain;
     char *version;
+    char *filename;
     version_data *base;
     void *data;
     const char *userdata_key = "mod_manager_init";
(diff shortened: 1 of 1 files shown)


jfclere:
@kimonides  Please merge the 4 commits into a single one.

kimonides:
> @kimonides Please merge the 4 commits into a single one.

Just finished merging the 4 commits into a single one.

jfclere:
run tests

Karm:
@jfclere The w2k12 box is dead until I can go to office and do something about it.  I have 2k19 up and running, although it is not fully geared for building openssl (missing perl). It should build httpd and mod_cluster just fine though. Lemme try...

PR #48, opened 2022-08-04
[MODCLUSTER-764] optimize by reusing find_node_context_host results

aogburn:
https://issues.redhat.com/browse/MODCLUSTER-764
https://issues.redhat.com/browse/MODCLUSTER-757

### DIFF
--- a/native/common/common.c
+++ b/native/common/common.c
@@ -319,6 +319,11 @@ node_context *find_node_context_host(request_rec *r, proxy_balancer *balancer, c
     const char *uri = NULL;
     const char *luri = NULL;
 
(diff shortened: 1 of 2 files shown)


jfclere:
According to my tests that causes a regression in the "performances":
http://localhost:8000/webapp1-9499 (I have created a different app for each node but the same happens with only 2 webapps)
using CacheShareFor 2

with PR48: 
 50%   5069
main without the PR48:
   50%     12

aogburn:
Hmm, a difference that large seems odd if we're just removing find_node_context_host calls.  Did something else change between the builds in your test?

PR #52, opened 2022-11-18
Fix compile warnings

jajik:
Fix some compile warnings. Mainly those about unused arguments of functions, but also a few others such as uninitialized values or signed/unsigned comparison.


rhusar:
Needs a rebase.

PR #54, opened 2022-11-23
Fix compile warnings and wrong magic constant value.

jajik:
Fix for some compile warnings. Mainly those about unused arguments of functions, but also a few others such as uninitialized values or signed/unsigned comparison.

Fixes also a typo in a magic constant macro comparison that I missed on last PR.


rhusar:
run tests

PR #60, opened 2023-02-09
Fix declaration-after-statement error

jajik:
Fix an error with `-Werror=declaration-after-statement` that was introduced in the previous commit 5a42af51883c14e845c692d022f97525aaa7d2d2


jfclere (changes requested):
There still some  error: ISO C90 forbids mixed declarations and code...
likie mod_proxy_cluster.c, common/common.c etc

jajik:
Should be finally fixed.

PR #61, opened 2023-02-27
Fix code formatting inconsistency

jajik:
While going through the codebase I've noticed that there are some formatting inconsistencies. This PR tries to fix that.

It fixes things like different spacing for cycles, around operators, etc. There is no functional change.


rhusar (approved):
LGTM – definitely the style I prefer :-)

rhusar:
Just a side note: generally changes like these make back-porting to legacy support branches unnecessarily more involved. Having said that, the easiest is to keep the style consitent from day one (most of the code here originated prior to github/PR reviews/CI and that's how it came to be). Anyway, mod_proxy_cluster is a place to fix these.

PR #66, opened 2023-04-18
Test suite improvements

jajik:
I went through the test suite and made some changes. Mainly

* change formatting of `tests.sh` script, renamed some functions, and fixed bugs
* updated httpd container image used in tests (it was not working)
* updated README

There is space for more improvements, mainly separating individual tests to be able to run them separately, but that's something for the future.

### DIFF
--- a/test/native/Dockerfile
+++ b/test/native/Dockerfile
@@ -1,10 +1,10 @@
-FROM tomcat:9.0
-ARG VERSION=1.4.1
+FROM tomcat:8.5
 
 WORKDIR /usr/local/tomcat
 
 COPY target/dependency/*.jar ./lib/
 COPY ./files/server.xml ./conf/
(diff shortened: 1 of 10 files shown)


jfclere (commented):
Why are you using 8.5?

jfclere (commented):
sed -i -e 's/LogLevel warn/LogLevel debug/g' /usr/local/apache2/conf/httpd.conf

Please Don't :D

jajik:
Good catch. I had an issue with Tomcat 9, those jsp files tests result in 404, older tomcat worked fine. (And then forgot about it. I'll have a look.)

> Please Don't :D

I found it useful for an image dedicated to testing, but ok, I'll remove it.

jfclere (approved):

rhusar (approved):
Generally looks good to me.

PR #67, opened 2023-04-18
Make watchdog logic threaded when possible

jajik:
Implement health checks using threads when available in a similar way mod_proxy_hcheck.

ref: https://issues.redhat.com/browse/MODCLUSTER-781


jfclere (changes requested):
Why MC_USE_THREADS? in fact we can use APR_HAS_THREADS directly (and bailout if that we don't have APR_HAS_THREADS (that mod_proxy_cluster won't work correctly without it).

jfclere:
We also need to document ModProxyClusterThreadCount

jajik:
> Why MC_USE_THREADS? in fact we can use APR_HAS_THREADS directly (and bailout if that we don't have APR_HAS_THREADS (that mod_proxy_cluster won't work correctly without it).

The advantage is that we get a possibility of not using threads (falling back to the previous behaviour). This is the same result as we get on system without APR_HAS_THREADS (i.e., it should work as it does now).

It seems to me that getting better granularity with one more macro is a better way of doing it, but I can remove it if it's a problem.

jfclere:
That is it not a problem.

PR #69, opened 2023-04-21
Use ap_hook_child_stopping() to allow cleaner stop of the watchdog logic + arrange thread logic + use more standard lock()/unlock()

jfclere:
replace the file/thread locks by apr_global_mutex_t.
arrange the thread logic to use it only in the blocking part of the health check logic.
basically the watchdog logic prevent the graceful restart of the child processes,

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -121,10 +121,10 @@ typedef struct version_data {
 } version_data;
 
 /* mutex and lock for tables/slotmen */
-static apr_thread_mutex_t *nodes_global_mutex = NULL;
-static apr_file_t *nodes_global_lock = NULL;
-static apr_thread_mutex_t *contexts_global_mutex = NULL;
-static apr_file_t *contexts_global_lock = NULL;
+static apr_global_mutex_t *node_mutex;
+static apr_global_mutex_t *context_mutex;
+static const char *node_mutex_type = "node-shm";
+static const char *context_mutex_type = "context-shm";
 
 /* counter for the version (nodes) */
 static apr_shm_t *versionipc_shm = NULL;
@@ -264,29 +264,13 @@ static int loc_worker_nodes_are_updated(void *data, unsigned int last)
     mconf->tableversion = last;
     return (0);
 }
-static apr_status_t lock_memory(apr_file_t *file, apr_thread_mutex_t *mutex)
-{
-    apr_status_t rv;
-    rv = apr_thread_mutex_lock(mutex);
-    if (rv != APR_SUCCESS)
-        return rv;
-    rv = apr_file_lock(file, APR_FLOCK_EXCLUSIVE);
-    if (rv != APR_SUCCESS)
-        apr_thread_mutex_unlock(mutex);
-    return rv;
-}
-static apr_status_t unlock_memory(apr_file_t *file, apr_thread_mutex_t *mutex)
-{
-    apr_file_unlock(file);
-    return(apr_thread_mutex_unlock(mutex));
-}
 static apr_status_t loc_lock_nodes(void)
 {
-    return(lock_memory(nodes_global_lock, nodes_global_mutex));
+    return(apr_global_mutex_lock(node_mutex));
 }
 static apr_status_t loc_unlock_nodes(void)
 {
-    return(unlock_memory(nodes_global_lock, nodes_global_mutex));
+    return(apr_global_mutex_unlock(node_mutex));
 }
 static int loc_get_max_size_context(void)
 {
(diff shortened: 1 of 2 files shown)


native/mod_manager/mod_manager.c
```diff
@@ -723,6 +695,16 @@ static int manager_init(apr_pool_t *p, apr_pool_t *plog,
      */
     mc_initialize_cleanup(p);
 
+    /* Create global mutex */
+    if (ap_global_mutex_create(&node_mutex, NULL, node_mutex_type, NULL, s, p, 0) != APR_SUCCESS) {
+        ap_log_error(APLOG_MARK, APLOG_ERR|APLOG_NOERRNO, 0, s,"manager_init: ap_global_mutex_create failed");
+        return !OK;
+    }
+    if (ap_global_mutex_create(&context_mutex, NULL, context_mutex_type, NULL, s, p, 0) != APR_SUCCESS) {
+        ap_log_error(APLOG_MARK, APLOG_ERR|APLOG_NOERRNO, 0, s,"manager_init: ap_global_mutex_create failed");
```
jajik, line 704: I would consider adding which mutex failed (node/context) as the text is the same for both. The same applies for the `apr_global_mutex_child_init`.

native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -1257,6 +1265,78 @@ static apr_status_t read_node_worker(int id, nodeinfo_t **node, proxy_worker *wo
...
+            ou->updatetime = now;
+        } 
+    } else {
+        ou->mess.num_failure_idle = 0;
+    }
+
+    node_storage->unlock_nodes();
+    return APR_SUCCESS;
+}
+
```
jajik, line 1339: I miss deallocation of the allocated pool. Is it because we rely on the (parent) pool deallocation?

native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -1277,7 +1357,14 @@ static void update_workers_lbstatus(proxy_server_conf *conf, apr_pool_t *pool, s
     /* update lbstatus if needed */
     for (i = 0; i < size; i++) {
         nodeinfo_t *ou;
+
```
jajik, line 1360: Some unnecessary spaces? (or github does not display changes correctly)

jajik (commented):
Could you also sync the fork please? (There is the `name` x `name_ex` issue)

native/mod_manager/mod_manager.c
```diff
@@ -3322,15 +3304,18 @@ static void  manager_child_init(apr_pool_t *p, server_rec *s)
...
-    if (apr_thread_mutex_create(&nodes_global_mutex, APR_THREAD_MUTEX_DEFAULT, p) != APR_SUCCESS) {
-        ap_log_error(APLOG_MARK, APLOG_ERR|APLOG_NOERRNO, 0, s,
-                    "manager_child_init: apr_thread_mutex_create failed");
-        return;
+
+    if (apr_global_mutex_child_init(&node_mutex, apr_global_mutex_lockfile(node_mutex), p) != APR_SUCCESS) {
+        ap_log_error(APLOG_MARK, APLOG_CRIT, 0, s, APLOGNO(02994)
+                     "Failed to reopen mutex %s in child",
+                     node_mutex_type);
+        exit(1);
```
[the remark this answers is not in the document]
jajik: Ok, so it seems it should work. (And `httpd` has some exit(1) too.) However we might consider using `EXIT_SUCCESS` and `EXIT_FAILURE` instead of `0`/`1`.

native/mod_proxy_cluster/mod_proxy_cluster.c
jfclere: yes we rely on the parent pool deallocation.

native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -2280,37 +2367,20 @@ static apr_status_t mc_watchdog_callback(int state, void *data,
                 mc_thread_pool = NULL;
                 ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, s, "Threads are not used");
             }
+            ap_log_error(APLOG_MARK, APLOG_ERR, 0, s, "mc_watchdog_callback, apr_thread_pool_create created!!!");
```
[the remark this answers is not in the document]
jfclere: not sure what is the problem... It looks like I have forgotten to remove the ap_log_error() here, but is there something else?

native/mod_proxy_cluster/mod_proxy_cluster.c
jajik: In this case, the parent pool is the `server->process->pool`. How often does it get deallocated, couldn't it be a problem? (I don't know.)

jfclere: OK so you are right the server->process->pool is not deallocated for whole life time of the child process...

native/mod_manager/mod_manager.c
jfclere: apr_global_mutex_child_init already log the node_mutex_type or context_mutex_type

jajik (approved):
It seems ok to me. Some things will be tackled with code styles in the future. Compilation on Windows is ok and tests are passing too.

PR #73, opened 2023-04-27
Add find_node_byhostport() and use it to find the nodes that are being removed.

jfclere:
The proxy logic uses host and port, we probably  need to use it more in mod_proxy_cluster to improve the reuse logic.

### DIFF
--- a/native/mod_manager/node.c
+++ b/native/mod_manager/node.c
@@ -275,3 +275,39 @@ mem_t * create_mem_node(char *string, int *num, int persist, apr_pool_t *p, slot
 {
     return(create_attach_mem_node(string, num, CREATE_SLOTMEM|persist, p, storage));
 }
+
+
+/**
+ * find the first node that corresponds to host/port.
+ * @param pointer to the shared table.
+ * @param node node to read from the shared table.
+ * @param host string containing the host.
+ * @param port string containing the port .
+ * @return APR_SUCCESS if all went well
+ */
+static apr_status_t loc_read_node_byhostport(void* mem, void **data, int id, apr_pool_t *pool) {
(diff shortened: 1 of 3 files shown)


native/mod_manager/node.c
```diff
@@ -275,3 +275,37 @@ mem_t * create_mem_node(char *string, int *num, int persist, apr_pool_t *p, slot
 {
     return(create_attach_mem_node(string, num, CREATE_SLOTMEM|persist, p, storage));
 }
+
+
+/**
+ * find the first node correspond to host/port.
+ * @param pointer to the shared table.
+ * @param node node to read from the shared table.
+ * @return address of the read node or NULL if error.
```
jajik, line 284: That's misleading. Better would be 

> @return `APR_SUCCESS` in case the node is found and its address is in `data`, `APR_NOTFOUND` otherwise

or something like that

jajik (changes requested):
Other than the one note, it seems ok to me.

native/mod_manager/node.c
jfclere: fixed

native/mod_manager/node.c
```diff
@@ -275,3 +275,39 @@ mem_t * create_mem_node(char *string, int *num, int persist, apr_pool_t *p, slot
...
 }
+
+
+/**
+ * find the first node that corresponds to host/port.
+ * @param pointer to the shared table.
+ * @param node node to read from the shared table.
+ * @param host string containing the host.
+ * @param port string containing the port .
+ * return APR_SUCCESS if all went well
```
rhusar, line 286: s/return/@return

jfclere: fixed thanks

PR #74, opened 2023-05-02
Apply style guide similar to httpd's one, tweak code style

jajik:
Related to: https://issues.redhat.com/browse/MODCLUSTER-783

I used `indent` with following options: `-i4 -npsl -di0 -br -nce -d0 -cli0 -npcs -nfc1 -nut -ncs -l120`.

The most of the changes are by `indent` itself, however, I removed some unnecessary parenthesis, inconsistent brackets, and made minor tweaks.

Please, I welcome any feedback.

### DIFF
--- a/native/.clang-format
+++ b/native/.clang-format
@@ -1,0 +1,34 @@
+Language: Cpp
+IndentWidth: 4
+PointerAlignment: Right
+ColumnLimit: 120
+AlignConsecutiveMacros: AcrossEmptyLinesAndComments
+AllowAllArgumentsOnNextLine: true
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+BreakBeforeBraces: Custom
+BraceWrapping:
+  BeforeElse: true
+  AfterFunction: true
+  AfterStruct: true
+  AfterUnion: true
+IndentCaseLabels: false
+IndentExternBlock: false
+MaxEmptyLinesToKeep: 2
+QualifierAlignment: Left
+SpaceAfterCStyleCast: false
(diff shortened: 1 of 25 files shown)


rhusar (approved):
Generally +1 from me – there is never ideal time to do this – but I don't think there will ever be a better time to do this other than now in mod_proxy_cluster. But final word is @jfclere 's.

rhusar:
BTW @jajik  Is there a some linter / code style checker tool that we can use to keep the style in place?

jajik:
Ok. I will have a look at settings we could use.

jajik:
@rhusar I added a `.clang-format` file for `clang-format` tool. To see changes it produces, I ran (from native directory):

```sh
for f in */*.c; do clang-format $f > $f-diff; done
```

and then used `meld` on corresponding files.

Also, see the trailing comma I added with the last commit. Without it, the format is different (and uglier) and unfortunately this seems like the best way to force this format (we could use an empty comment too, but that seems like a worse option) . Other than that, I think it is usable but it differs from `indent` slightly.

I am going to see all the changes it proposed and tweak it, if necessary. I'll be happy for any comment.

jajik:
**Update 3**:
I changed a few bits in the proposed `.clang-format`. I also applied the rules to the code base in [the latest commit](b8191d76dd1c7f2d6ed6c27e1188f335ce5e849c). Overall, I think that the results are better than with `indent`, even though I have some reservations about (some) comments' formatting and (some) array initializer lists. For the latter, I added `clang-format off/on` comments to preserve the format we already have.

jajik:
Done.

PR #75, opened 2023-05-02
CI: add a smoke CMake build on Fedora 38

rhusar:
With an increased turnaround, let's make sure things don't break unintentionally with a CMake smoke build on Fedora 38 for a start.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,36 @@
+name: CI
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - '*'
+jobs:
+  ci:
+    name: cmake-fedora-latest
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


jajik:
That's a great idea! I would suggest adding a second pipeline/task that will use the following (from native directory):

```sh
for m in advertise/ mod_*; do
    echo $m
    cd $m
    sh buildconf
    ./configure --with-apxs=$APACHE_DIR/bin/apxs
    make
    cd ..
done
```

The reason being different command line arguments for the compiler (namely `c89` and some `missing-prototypes`, `missing-declarations`, etc.). But it requires httpd.

rhusar:
@jajik Thanks - added as a separate job (this way they will run concurrently, so if only one build if broken in a PR it's easier and faster to debug).

rhusar:
OK thanks - I will merge this now so we can see how reliably this works.

rhusar:
That looks good now.

image

I have disabled the Jenkins CI for now until it gets fixed.

jajik:
That looks much better! :D

PR #76, opened 2023-05-03
Set MaxConnectionsPerChild to 16 for tests

jajik:
With recent changes, for tests it is necessary to set MaxConnectionsPerChild so that nodes get removed.

### DIFF
--- a/test/native/httpd/install.sh
+++ b/test/native/httpd/install.sh
@@ -49,6 +49,12 @@ if [ -f /tmp/$FILECONF ]; then
   echo "Include conf/$FILECONF" >> /usr/local/apache2/conf/httpd.conf
 fi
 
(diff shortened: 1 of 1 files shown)


test/native/httpd/install.sh
```diff
@@ -49,6 +49,8 @@ if [ -f /tmp/$FILECONF ]; then
   echo "Include conf/$FILECONF" >> /usr/local/apache2/conf/httpd.conf
 fi
 
+echo -e "\nMaxConnectionsPerChild    16"  >> /usr/local/apache2/conf/httpd.conf
```
rhusar, line 52: Perhaps add a note for the future why this is required here?

jajik: I added a rather lengthy note. Is it ok like that?

rhusar: Looks good, thanks

jfclere (approved):

PR #79, opened 2023-05-09
Improve debugging when using ProxyPass

jfclere:
Improve debugging when using ProxyPass.
return 404 instead 503 when no context corresponding to the ProxyPass.


jajik (approved):
Seems ok to me.

PR #81, opened 2023-05-10
Add build with httpd in dev mode

jajik:
There is difference in compilation between regular `make`/`cmake` and the same with `httpd` configured with `--enable-maintainer-mode` and some additional flags. For example, using C++ style comments is ok unless `httpd` is configured as stated above.

Therefore, I think it would be worth creating an image with httpd configured and compiled.

This Draft PR contains a possible solution (using `quay.io/vchlup/httpd-dev`); it is a draft because I think the image should be within the modcluster namespace/organization, not in mine.

The `Containerfile` I used for the image:

```Containerfile
FROM ubi9/ubi:latest
# install necessary tools (+ some useful ones too)
# RUN yum module enable subversion
RUN dnf update -y && dnf install -y git gcc make libtool python3 autoconf libxml2-devel pcre2-devel
# libxml workaround
RUN cd /usr/include/ && ln -s libxml2/libxml .
# get httpd and checkout the appropriate version
RUN git clone https://github.com/apache/httpd
WORKDIR /httpd/
RUN git checkout 2.4.57
# get apr library
RUN git clone https://github.com/apache/apr srclib/apr
# configure httpd
RUN export CFLAGS="-g"
# define APACHE_DIR variable
RUN echo "APACHE_DIR=/usr/local/apache2/" >> /root/.bashrc
RUN ./buildconf
RUN ./configure --prefix=/usr/local/apache2 --with-included-apr --enable-proxy-ajp --enable-maintainer-mode \
                --enable-so --enable-proxy --enable-proxy-http --enable-proxy-wstunned --enable-proxy-hcheck \
                --with-port=8000 --with-libxml2
# build httpd
RUN make
RUN make install
# start bash
CMD ["/bin/bash", "-c", "bash"]
```

The image is then created as usual with `podman build -t quay.io/namespace/image:tag`. After pushing, you have to allow reading (simply making it public seems the best option).

In my forked repository you can see the proposed job in action: [passed](https://github.com/jajik/mod_proxy_cluster/actions/runs/4945494403/jobs/8842400460), [failed](https://github.com/jajik/mod_proxy_cluster/actions/runs/4945560068/jobs/8842449631).

**Updated**: Now we're using ubi9 and `quay.io`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,4 +93,24 @@ jobs:
             rm diff.txt || true # we don't fail if the file does not exist
           done;
           exit $code
+  make-httpd-maintainer-mode:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/mod_cluster/ci-httpd-dev
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Build
+        run: |
+          cd native
+          for module in advertise/ mod_*; do \
+            cd $module; \
+            sh buildconf; \
+            ./configure  --with-apxs=/usr/local/apache2/bin/apxs; \
+            make || exit 1; \
+            cd ..; \
+          done;
 
--- a/native/mod_proxy_cluster/mod_proxy_cluster.c
+++ b/native/mod_proxy_cluster/mod_proxy_cluster.c
@@ -1261,8 +1261,6 @@ static void *APR_THREAD_FUNC check_proxy_worker(apr_thread_t *thread, void *data
     apr_pool_t *rrp;
     request_rec *rnew;
 
-    (void)thread;
-
     watchdog_thread_args_t *targs = (watchdog_thread_args_t *)data;
     proxy_worker *worker = targs->worker;
     apr_pool_t *pool = targs->pool;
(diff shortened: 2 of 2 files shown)


jajik:
I've removed 994b0e6c33b40e50e7151a7416aacf52344ade51 as it is part of another PR (I'll rebase this after merge.)

.github/workflows/ci.yml
```diff
@@ -93,4 +93,25 @@ jobs:
             rm diff.txt || true # we don't fail if the file does not exist
           done;
           exit $code
+  make-httpd-maintainer-mode:
+    runs-on: ubuntu-latest
+    name: "Build with httpd in maintainer-mode"
+    container:
+      image: quay.io/vchlup/httpd-dev:latest
```
rhusar, line 100: Update for: https://quay.io/repository/mod_cluster/ci-httpd-dev

.github/workflows/ci.yml
```diff
@@ -93,4 +93,25 @@ jobs:
             rm diff.txt || true # we don't fail if the file does not exist
           done;
           exit $code
+  make-httpd-maintainer-mode:
+    runs-on: ubuntu-latest
+    name: "Build with httpd in maintainer-mode"
```
rhusar, line 98: If we are putting human readable names to jobs, then lets do it for all or none.

.github/workflows/ci.yml
rhusar: I did prefix this with "ci-" so that once we have images for general consumptions aren't confused with internal images - but let me know if there's a better name.

.github/workflows/ci.yml
jajik: Good point, removed.

jajik:
Changes are in, CI jobs passed -> converting to regular PR.

jfclere (approved):

rhusar (approved):
Looks good to me now.

Issue #63, opened 2023-03-28
Unreachable code in function with an empty return


jajik:
Hello,

I have found this strange `if .. else` statement in the code.

https://github.com/modcluster/mod_proxy_cluster/blob/11f6f73f06c8d1ff62c16b6b8c67b47fb20fd73f/native/mod_proxy_cluster/mod_proxy_cluster.c#L1268-L1271

The function has some additional statements that are therefore unreachable. Is this a bug, or is it intended? If the latter, wouldn't it be better to omit the function completely as it does not do what it says? Or am I missing something?

It was introduced here  8576d58d82ba667ac703030f3b1852eab40f9547.

Thanks.

rhusar:
Assigning @jfclere to clean this up. Thanks!

jfclere:
that is obviously a bug!

jfclere:
the:
+++
else 
     return; 
+++
Should be removed.

jajik:
The else statement was removed.

PR #83, opened 2023-05-17
Remove a dead code from mod_proxy_cluster

jajik:
I looked into #63, trying to remove the dead code. It seems that the whole function can be removed without affecting the functionality. Tests in the repository are passing. Also, the described behaviour in JIRA related to the fix that introduced the bug is still not reproducible with this change – the node gets flagged REMOVED with hung JVM (the actual removal then depends on `MaxConnectionsPerChild`).

Please, check that I am not missing something.


jfclere (commented):
I don't get why you removed the code...

jfclere (approved):

jajik:
> I don't get why you removed the code...

The logic was not in use for two years, so I assumed the current state is the one. Sorry about that.

----------------------

Anyway, I've dropped the previously proposed commit and removed the else branch as you suggested. Tests are passing and CI too, so I'll merge it now.

Issue #82, opened 2023-05-16
Make/Compilation Error


aramir:
On `make`,  geting:
```
[ 42%] Building C object mod_manager/CMakeFiles/mod_manager.dir/mod_manager.c.o
/opt/mod_cluster/mod_proxy_cluster-main/native/mod_manager/mod_manager.c: In function process_dump
/opt/mod_cluster/mod_proxy_cluster-main/native/mod_manager/mod_manager.c:1612:13: error: a label can only be part of a statement and a declaration is not a statement
             char *status;
             ^~~~
```
for this code block:
```c
case TEXT_XML:
    char *status;
    status = "REMOVED";
    switch (ou->status) {
    case ENABLED:
	status = "ENABLED";
	break;
    case DISABLED:
	status = "DISABLED";
	break;
    case STOPPED:
	status = "STOPPED";
	break;
    }
    ap_rprintf(r, "<Context id=\"%d\" path=\"%.*s\">\
		    <Vhost>%d</Vhost>\
		    <Node>%d</Node>\
		    <Status id=\"%d\">%s</Status>\
		   </Context>",
	       id[i], (int)sizeof(ou->context), ou->context, ou->vhost, ou->node, ou->status, status);
    break;
```
surrounding the case with `{}` fixed it

jajik:
Hello, could you please share which compiler and which flags were used?

jajik:
Great, thank you! The fix is ready in PR #84 along with a tweak for CI jobs.

[...]

PR #84, opened 2023-05-17
Fix switch/case label error, tweak status handling

jajik:
Fix a compilation error (#82) introduced with the new code style. The PR also refactors duplicated functionality.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,9 @@ on:
 jobs:
   cmake-fedora-latest:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        compiler: [gcc, clang]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


jajik:
I added a second commit that makes the pipeline to build with both, clang and gcc. It might prevent some compilation errors as the one in #82.

(Without the 1ebdbfe11b137e8c9e179a28ddf88c5fde1f0f05, the job would fail. See https://github.com/jajik/mod_proxy_cluster/actions/runs/5011990175/jobs/8983418382)

rhusar:
> Note that the CI reports `Waiting for status to be reported` for `cmake-fedora-latest` job. This is due to the change in CI, where the job of that name now runs in two flavours – the old name is no longer used, but PR was opened while it existed and it stayed there.
> 
> TL;DR it should be ignored.

That's correct - I will just change the required checks set it once we merge this.

rhusar:
> > Note that the CI reports `Waiting for status to be reported` for `cmake-fedora-latest` job. This is due to the change in CI, where the job of that name now runs in two flavours – the old name is no longer used, but PR was opened while it existed and it stayed there.
> > TL;DR it should be ignored.
> 
> That's correct - I will just change the required checks set once we merge this.

Required checks are now updated.

PR #85, opened 2023-05-17
Make httpd's install script fail if branch does not exist

jajik:
Our testing httpd image has  a variable `BRANCH` by which one can specify a source branch for mod_proxy_cluster. However, if a branch of the given name does not exist, `git checkout` command fails and tests continue on the default main branch. This can lead to false positive/negative tests results.

This PR changes the behaviour – the script now fails in case the checkout is not successful making the image stop.


rhusar (approved):

rhusar:
Merged, thanks @jajik

PR #89, opened 2023-05-25
CI: do not fail fast on matrix jobs.

rhusar:


rhusar:
If anyone wonders, here's the docs https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

PR #91, opened 2023-05-25
Add a note to README about applying styles suggestions

jajik:
ref #90


rhusar:
Merged, thanks @jajik

Issue #92, opened 2023-05-30
Correct style, unsigned int and use constant when possible. 


jfclere:
We are mixing
if (bla) { and if (bla)
int count, unsigned int count
some of the variable should be constant to make further development more easy.

jajik:
Just to clarify, do we mind using `if` statements without brackets? I think it's ok without brackets.

The issue I had in #88 was mixing `if`/`else` where one branch has brackets and the other one does not. E.g.,

```cpp
if (bla)
    // do something
else {
    // do something else
}
```

If we want to enforce brackets in all cases, we can, but I think it's unnecessary...

rhusar:
> Just to clarify, do we mind using `if` statements without brackets? I think it's ok without brackets.
> 
> The issue I had in #88 was mixing `if`/`else` where one branch has brackets and the other one does not. E.g.,
> 
> ```c++
> if (bla)
>     // do something
> else {
>     // do something else
> }
> ```
> 
> If we want to enforce brackets in all cases, we can, but I think it's unnecessary...

The code style https://github.com/modcluster/mod_cluster uses brackets always, I don't see the problem adopting the same in the httpd module code.

jajik:
We can enforce it, but due to the rule where closing bracket of `if` is on a different line than `else`, we'll get

```c
if (bla) {
    foo();
}
else {
    bar();
}
```

instead of

```c
if (bla)
    foo();
else
    bar();
```

But ok, I'll create a PR where I fix the above and add a rule to .clang-format.

jfclere:
I am like @rhusar: brackets everywhere  is better! let's enforce it!

jajik:
TODO:
- [ ] `if`, `else` brackets
- [ ] use `unsigned` for countrs 
- [ ] use `unsigned` instead of `unsigned  int`
- [ ] some of the variable should be constant to make further development more easy.

PR #96, opened 2023-06-02
Draft: Improve styles

jajik:
This PR should address #92 .


rhusar:
FYI – looking at the logs opened https://github.com/modcluster/mod_proxy_cluster/pull/98

rhusar (changes requested):
The version of clang-format we use on Ubuntu in CI doesn't support this option.

    .clang-format:15:1: error: unknown key 'InsertBraces'

jajik:
> The version of clang-format we use on Ubuntu in CI doesn't support this option.
> 
> ```
> .clang-format:15:1: error: unknown key 'InsertBraces'
> ```

Yes, I'm going to use a container here too.

rhusar:
> > The version of clang-format we use on Ubuntu in CI doesn't support this option.
> > ```
> > .clang-format:15:1: error: unknown key 'InsertBraces'
> > ```
> 
> Yes, I'm going to use a container here too.

Let's not over-complicate this - i ll open a PR to get the latest clang.

rhusar:
> > > The version of clang-format we use on Ubuntu in CI doesn't support this option.
> > > ```
> > > .clang-format:15:1: error: unknown key 'InsertBraces'
> > > ```
> > 
> > 
> > Yes, I'm going to use a container here too.
> 
> Let's not over-complicate this - i ll open a PR to get the latest clang.


This fixes the problem - https://github.com/modcluster/mod_proxy_cluster/pull/96#issuecomment-1576439772

rhusar:
Starting to look good now. @jfclere happy to agree on this style?

jajik:
I think it is ready to go. Regarding the linked issue #92, there is one thing missing – adding `const` where it makes sense. I am currently going through the code, but it will take some time and I think a separate PR is probably more appropriate.

rhusar:
@jfclere please approve. Thanks.

jfclere (approved):

jajik:
@rhusar Your approval is needed too. :)

native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -641,16 +642,20 @@ static proxy_balancer *add_balancer_node(nodeinfo_t *node, proxy_server_conf *co
...
+        if (balan == NULL) {
             return balancer; /* Done broken */
+        }
         /* StickySession, StickySessionRemove we hack it via the lbpname (16 bytes) */
-        if (!balan->StickySession)
+        if (!balan->StickySession) {
             strcpy(balancer->s->lbpname, MC_NOT_STICKY);
-        else
+        }
+        else {
```
rhusar, line 652: Hm, does the clang formatter require the `else {` to be on a new line?

I would expect the 'else' to be on the same line as the bracket of the previous block, i.e.:

`} else {`

That's the style we have always used for Java - [example](https://github.com/modcluster/mod_cluster/blob/5babe75797db30c334f099f42992db130b784108/container/tomcat-8.5/src/main/java/org/jboss/modcluster/container/tomcat/TomcatConnector.java#L74https://github.com/modcluster/mod_cluster/blob/5babe75797db30c334f099f42992db130b784108/container/tomcat-8.5/src/main/java/org/jboss/modcluster/container/tomcat/TomcatConnector.java#L74). Starting a new line with `else` hurts my brain.

rhusar (changes requested):

native/mod_proxy_cluster/mod_proxy_cluster.c
jajik: Yes, it's how they do it in httpd, see https://github.com/modcluster/mod_proxy_cluster/pull/74 and related https://issues.redhat.com/browse/MODCLUSTER-783.

In clang-format, it's the BraceWrapping option. (I hate it too. :smiley:)

I guess we can change it, but the decision is on @jfclere I guess. We can either stick to httpd style, or go with the mod_cluster/Java part style.

jfclere: We can always arrange the style later if one day we give the mod_proxy_cluster the Apache httpd.
So we could with the mod_cluster/Java part style for the moment.

jajik: I changed the style in the latest commit.

rhusar (approved):
I think this looks pretty good now. 

I have looked into https://issues.redhat.com/browse/MODCLUSTER-783 and the linked apache style from 1996 and it's safe to say that it's out of style today (pun intended).

rhusar:
Thanks @jajik @jfclere !

PR #100, opened 2023-06-06
Add Windows runner into CI

jajik:
In the past it occurred multiple times that a change was introduced that worked on Linux, but broke build on Windows. Now when we no longer use CI via Jenkins, we have limited ability to catch those changes. This PR adds windows runner (`cmake` using) into the CI.

Also, there is a fix that moves an optional function declaration into the correct file (it caused compilation on windows to fail).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,4 +120,20 @@ jobs:
             make || exit 1; \
             cd ..; \
           done;
-
+  cmake-windows-latest:
+    runs-on: windows-latest
+    env:
+      APACHE_LOUNGE_DISTRO_VERSION: 2.4.57
+      HTTPD_DEV_HOME: '${{ github.workspace }}\httpd-apache-lounge\Apache24'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Developer Command Prompt for Microsoft Visual C++
+        uses: ilammy/msvc-dev-cmd@v1
+      - name: Get httpd
+        run: |
+          curl.exe --output "httpd-apache-lounge.zip" --url "https://www.apachelounge.com/download/VS16/binaries/httpd-${{ env.APACHE_LOUNGE_DISTRO_VERSION }}-win64-VS16.zip"
+          Expand-Archive -Path "httpd-apache-lounge.zip" -DestinationPath "httpd-apache-lounge"
+      - name: Build
+        run: |
+          ./native/scripts/windows-build.bat
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ native/*/configure
 test/java/node*/
 bin/
 target/
+
+# Folder usually used for CMake
+**/build
--- a/native/include/mod_proxy_cluster.h
+++ b/native/include/mod_proxy_cluster.h
@@ -30,8 +30,6 @@
 
 #define MOD_CLUSTER_EXPOSED_VERSION "mod_cluster/2.0.0.Alpha1-SNAPSHOT"
 
-APR_DECLARE_OPTIONAL_FN(apr_status_t, balancer_manage, (request_rec * r, apr_table_t *params));
-
 struct balancer_method
 {
     /**
--- a/native/mod_manager/mod_manager.h
+++ b/native/mod_manager/mod_manager.h
@@ -25,6 +25,10 @@
  * @version $Revision$
  */
 
+#include "apr_optional.h"
+
+APR_DECLARE_OPTIONAL_FN(apr_status_t, balancer_manage, (request_rec *, apr_table_t *));
+
 struct mem
 {
     ap_slotmem_instance_t *slotmem;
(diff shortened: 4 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -120,4 +120,20 @@ jobs:
...
+  cmake-windows-latest:
+    runs-on: windows-latest
+    env:
+      APACHE_LOUNGE_DISTRO_VERSION: 2.4.57
+      WORKSPACE: 'D:\a\mod_proxy_cluster\mod_proxy_cluster\'
+      HTTPD_DEV_HOME: 'D:\a\mod_proxy_cluster\mod_proxy_cluster\httpd-apache-lounge\Apache24'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - uses: ilammy/msvc-dev-cmd@v1
```
rhusar, line 132: Let's add a human readable name for this step, e.g. "Setup Developer Command Prompt for Microsoft Visual C++"

.github/workflows/windows-build.bat
```diff
@@ -0,0 +1,44 @@
+REM kudos to Michal Karm Babacek <karm@fedoraproject.org>
+REM this script is from (https://github.com/modcluster/ci.modcluster.io)
```
rhusar, line 2: Is this script 1:1 with version in `ci.modcluster.io`? In that case let's just check out that repo as well, rather than repeat / copy the file.

rhusar (changes requested):

.github/workflows/ci.yml
jajik: Done, thank you.

.github/workflows/ci.yml
```diff
@@ -120,4 +120,21 @@ jobs:
...
+        uses: actions/checkout@v3
+      - name: Setup Developer Command Prompt for Microsoft Visual C++
+        uses: ilammy/msvc-dev-cmd@v1
+      - name: Get httpd
+        run: |
+          curl.exe --output "httpd-2.4.57-win64-VS16.zip" --url "https://www.apachelounge.com/download/VS16/binaries/httpd-2.4.57-win64-VS16.zip"
+          Expand-Archive -Path "httpd-2.4.57-win64-VS16.zip" -DestinationPath "httpd-apache-lounge"
+      - name: Build
+        run: |
+          ./.github/workflows/windows-build.bat
```
rhusar, line 140: Am I wrong to assume this build file is usable outside of GitHub Actions? I think this should be in `/native` rather than in `.github/`.

rhusar (changes requested):

.github/workflows/ci.yml
jajik: Done, it should be a bit cleaner now.

.github/workflows/ci.yml
```diff
@@ -120,4 +120,20 @@ jobs:
...
+      HTTPD_DEV_HOME: '${{ github.workspace }}\httpd-apache-lounge\Apache24'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Developer Command Prompt for Microsoft Visual C++
+        uses: ilammy/msvc-dev-cmd@v1
+      - name: Get httpd
+        run: |
+          curl.exe --output "httpd-2.4.57-win64-VS16.zip" --url "https://www.apachelounge.com/download/VS16/binaries/httpd-2.4.57-win64-VS16.zip"
+          Expand-Archive -Path "httpd-2.4.57-win64-VS16.zip" -DestinationPath "httpd-apache-lounge"
```
rhusar, line 136: Can we make these to use the env variable `APACHE_LOUNGE_DISTRO_VERSION`? 3x

rhusar (changes requested):

.github/workflows/ci.yml
rhusar, line 136: Perhaps like this? Not a windows expert :)

```suggestion
          curl.exe --output "httpd-%APACHE_LOUNGE_DISTRO_VERSION%-win64-VS16.zip" --url "https://www.apachelounge.com/download/VS16/binaries/httpd-%APACHE_LOUNGE_DISTRO_VERSION%-win64-VS16.zip"
          Expand-Archive -Path "httpd-%APACHE_LOUNGE_DISTRO_VERSION%-win64-VS16.zip" -DestinationPath "httpd-apache-lounge"
```

rhusar: BTW If that has a chance of working you can press "Commit suggestion"

.github/workflows/ci.yml
rhusar: Perfect

.github/workflows/ci.yml
jajik: I tried it before, but it did not work. However it might be a problem (like a typo) on my side now that I'm thinking about it... let's wait and see. (I tried the `commit changes`, but even if it passes, I'll squash it.)

Good catch, thank you.

rhusar: > That didn't work, but `"${{ env.APACHE_LOUNGE_DISTRO_VERSION }}"` did. Also, there is no reason to name the zip locally with such a long name -> thank you, it's even better now.

Good to now - sounds like github is resolving the expression before passing onto windows.

rhusar (approved):
Looks good to me now.

rhusar:
Great work, thanks @jajik !

PR #101, opened 2023-06-06
Remove unused out-of-date file.

rhusar:


jajik:
Does it make sense to remove other files too? Namely

* `local.properties`
* `test-jars.xml`
* `build-test.xml`

If we want to go big, we could remove the whole `test/java` part and then flatten the structure from `test/native` to `test`.

rhusar:
> Does it make sense to remove other files too? Namely
> 
> * `local.properties`
> * `test-jars.xml`
> * `build-test.xml`
> 
> If we want to go big, we could remove the whole `test/java` part and then flatten the structure from `test/native` to `test`.

We should have a discussion for cleaning  those up or updating. But this deletion is out of that scope, these instructions are purely out of date and useless.

jajik:
Ok.

rhusar:
Thanks @jajik

PR #106, opened 2023-06-15
MODCLUSTER-674 Add Contributor Code of Conduct (Contributor Covenant …

rhusar:
…v2.1)

Add Contributor Covenant Code of Conduct to this repository as well.

Synchronize with https://github.com/modcluster/mod_cluster/blob/main/CODE_OF_CONDUCT.md

BTW while it's possible to add some of these files in a special github repository, I generally recommend not using this approach, since those files won't be available in people's repos clones.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,132 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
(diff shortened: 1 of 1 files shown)


jfclere (approved):

rhusar:
Thanks @jfclere

Issue #102, opened 2023-06-08
Refactor long functions


jajik:
There are some very long functions (e.g., `create_worker`); refactor them into smaller ones.

PR #107, opened 2023-06-20
Refactor create_worker, minor code tweaks

jajik:
Addresses #102 . There are still functions worthy of attention (hence the Draft marking) such as `internal_find_best_byrequests` or `update_workers_lbstatus`.

### DIFF
--- a/native/mod_proxy_cluster/mod_proxy_cluster.c
+++ b/native/mod_proxy_cluster/mod_proxy_cluster.c
@@ -1307,6 +1307,109 @@ static void *APR_THREAD_FUNC check_proxy_worker(apr_thread_t *thread, void *data
 }
 
 /*
+ * Returns 1 if the caller function should continue processing.
+ */
+static int internal_update_lbstatus(proxy_server_conf *conf, apr_pool_t *pool, server_rec *server, apr_time_t now,
+                                    nodeinfo_t *ou, int id, const proxy_worker_shared *stat)
+{
+    char sport[7];
+    watchdog_thread_args_t targs;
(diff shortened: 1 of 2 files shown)


jajik:
Marking ready to review.

I tried to split three very big functions (`create_worker`, `internal_find_best_byrequests`, `update_workers_lbstatus`) into smaller ones and although there is still room for improvements, I think we can go with this. I also removed some unnecessary else branches after return.

jajik:
(I've just rebased to the latest main.)

native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -1325,6 +1428,7 @@ static void update_workers_lbstatus(proxy_server_conf *conf, apr_pool_t *pool, s
     /* update lbstatus if needed */
     for (i = 0; i < size; i++) {
         nodeinfo_t *ou;
+        int cont;
```
jfclere, line 1431: Why here? and not in the if (child_stopping) { }?

jfclere (approved):

native/mod_proxy_cluster/mod_proxy_cluster.c
jfclere: if () {
    int count;
    ....
}
is OK

if () {
    ...
    int count;
   ...
}
is NOT.

jajik: Well, it turned out I was wrong :) Fixed, thank you @jfclere !

PR #108, opened 2023-06-21
tests: Set IMG variable to default if not defined

jajik:
Add a default value for IMG variable so that we are a bit more consistent (also, it is annoying to have to define it every time if one sticks to the recommended value).

### DIFF
--- a/test/native/tests.sh
+++ b/test/native/tests.sh
@@ -11,8 +11,12 @@ fi
 if [ -z ${ITERATION_COUNT+x} ]; then
     ITERATION_COUNT=50 # the number of iteration of starting/stopping a tomcat
 fi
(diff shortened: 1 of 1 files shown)


rhusar (approved):

rhusar:
This whole thing probably needs a bit of a redesign (i.e. i am not sure why one would use quay for running tests locally?) but let's get back to it later.

jajik:
Yes, it would be good to rewrite it completely and make it more modular. It would be also nice to run some reasonable subset in CI, but that's for the future. Thanks for the merge.

rhusar:
> Yes, it would be good to rewrite it completely and make it more modular. It would be also nice to run some reasonable subset in CI, but that's for the future. Thanks for the merge.

That's the plan!

PR #109, opened 2023-06-22
Add a test for MODCLUSTER-785

jfclere:
New and make script more flexible.


test/native/MODCLUSTER-785/testit.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+source ../includes/script.bash
+
+# check IMG...
+if [ -z ${IMG} ]; then
+  echo "IMG needs to defined, please try"
+  echo "export IMG=quay.io/${USER}/tomcat_mod_cluster"
+  exit 1
+fi
```
jajik, line 9: Don't we want to include these tests in the global testsuite too?

Then with the [latest](https://github.com/modcluster/mod_proxy_cluster/commit/1625f7846fa42bc520bd73738317e47fa1fad1db) changes this would not be needed. I not I would suggest to use the same approach (i.e., if not defined, we should go with the default and inform the user).

test/native/MODCLUSTER-785/testit.sh
```diff
@@ -0,0 +1,128 @@
...
+removetomcats
+
+# and httpd
+docker stop MODCLUSTER-785
+docker rm MODCLUSTER-785
+
+# build httpd + mod_proxy_cluster
+#
+LOCREPO=`git remote -v | grep origin | grep fetch | awk ' { print $2 } ' | awk -F/ ' { print $4 } '`
+REPO="https://raw.githubusercontent.com/modcluster/mod_proxy_cluster/main/test/native/MODCLUSTER-785/mod_proxy_cluster.conf"
```
jajik, line 30: I don't like those names. `LOCREPO` is more `ORIGINREPO` or `REPOORIGIN` or something like that, and `REPO` is not repository at all. (Maybe `MPCCONF` or `REMOTECONF`?).

test/native/MODCLUSTER-785/testit.sh
```diff
@@ -0,0 +1,128 @@
...
+if [ $LOCREPO != "modcluster" ]; then
+  BRANCH=`git branch -v | awk ' { print $2 } '`
+  REPO="https://raw.githubusercontent.com/$LOCREPO/mod_proxy_cluster/${BRANCH}/test/native/MODCLUSTER-785/mod_proxy_cluster.conf"
+fi
+rm -f nohup.out
+nohup docker run --network=host -e HTTPD=https://dlcdn.apache.org/httpd/httpd-2.4.57.tar.gz -e SOURCES=${SOURCES} -e BRANCH=${MDBRANCH} -e CONF=$REPO --name MODCLUSTER-785 quay.io/${USER}/mod_cluster_httpd &
```
jajik, line 36: There is no reason supply `HTTPD` variable when it is the same is the default. Also, wouldn't it be better to use `-d` instead of `&`.?

test/native/MODCLUSTER-785/testit.sh
```diff
@@ -0,0 +1,128 @@
...
+    exit 1
+  fi
+  i=`expr $i + 1`
+  if [ $i -gt 60 ]; then
+    break
+  fi
+  echo "*${http_code}*"
+done
+
+echo "MODCLUSTER-785 Done!"
```
jajik, line 128: It would be good to stop and remove the running containers too.

jajik (commented):
I have a few notes.

I think we should set up the global test script to run these as well. I'll look into that once we merge this one.

test/native/MODCLUSTER-785/testit.sh
jfclere: I am bad with names... I will fix that.

test/native/MODCLUSTER-785/testit.sh
jfclere: I will fix that.. But it is probably everywhere...

test/native/MODCLUSTER-785/testit.sh
jfclere: Do we want to do that if the test fails? I guess no ;-)

test/native/MODCLUSTER-785/testit.sh
jfclere: I don't get the HTTPD variable part...

jajik: If you look into https://github.com/modcluster/mod_proxy_cluster/blob/main/test/native/httpd/Dockerfile, there is `ENV HTTPD=https://dlcdn.apache.org/httpd/httpd-2.4.57.tar.gz`, so you don't have to run it with `-e HTTPD=https://dlcdn.apache.org/httpd/httpd-2.4.57.tar.gz`.

test/native/MODCLUSTER-785/testit.sh
jajik: I think we can let it running if it fails so that it's easy to check the logs.

rhusar: Not cleaning up after running tests is an antipattern. Ideally, let this be controled by an optional 'debug' option.

rhusar: Makes sense (please file an issue to track that), thanks.

test/native/MODCLUSTER-785/testit.sh
jfclere: OK we might need HTTP as an environment variable to test old version of httpd like 2.4.53 ;-)

jfclere: dlcdn.apache.org only has only the latest version normally.

jfclere: I am also thinking of an option to run with a locally install httpd, which is what I need to debug MODCLUSTER-785.

jajik: If it is due to the maintainer mode, you could try the image we run in CI. https://github.com/modcluster/ci.modcluster.io/tree/main/httpd-dev

jfclere: Actually the -d option is a BAD idea... the waitforhttpd waits for the nohup file...

test/native/MODCLUSTER-785/testit.sh
```diff
@@ -0,0 +1,132 @@
+#!/bin/bash
+source ../includes/script.bash
+
+# USe default if IMG not defined.
+if [ -z ${IMG+x} ]; then
+    IMG=IMG=quay.io/$USER/tomcat_mod_cluster
```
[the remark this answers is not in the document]
jfclere: Oops fixed.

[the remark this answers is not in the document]
rhusar: BTW @jfclere please squash the 'oops' commit into the other commits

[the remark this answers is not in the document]
jfclere: Done

rhusar:
I would suggest to get into the habit of rebasing the branches before pusing - there is a reminder for that on github "This branch is out-of-date with the base branch" - basically the tests would have been run with old revision.

jfclere:
@rhusar but the new test has to be run by hands for the moment, so that is OK for "this" PR.

jajik:
@jfclere I believe everything except the rebase was solved (maybe we could try Update with Rebase button here). It would be nice to merge it so that I can then incorporate it into #120. :)

jajik (approved):
Approve with regard to the previous comment.

rhusar (approved):

rhusar:
Thanks @jfclere and @jajik (and I am assuming you will cleanup some of the stuff from here in https://github.com/modcluster/mod_proxy_cluster/pull/120 ;-) ).

PR #113, opened 2023-06-26
Fix wrong variable definition in tests

jajik:
In the last improvement of tests I made an error and the variable definition included its name. This fixes that.


rhusar (approved):

rhusar:
Thanks @jajik

PR #114, opened 2023-06-27
Deduplication

jajik:
PR moves `loc_get_id` into `common.c`. This solution uses the id position within the corresponding structs (at the beginning), which can be a bit tricky. The alternative solution with `common.h` and macro controlled getter was explored as well, but it was way too complicated (we use both `make` and `cmake` + Windows behave differently than sane OSs).

I'm not sure we want to merge this – maybe it would be better to accept some duplication so I'm opening this as a draft.


jfclere (changes requested):
Please put that in another PR.

native/include/node.h
```diff
@@ -51,6 +51,16 @@ typedef struct mem mem_t;
 /* configuration of the node received from jboss cluster. */
 struct nodemess
 {
+    /* part updated in httpd */
+    int id;                  /* id in table and worker id */
```
jfclere, line 55: The comment needs to after the int id;
In fact please just move the int id;

jfclere (changes requested):

native/include/node.h
jajik: I wanted to preserve the structure within the struct, but as you wish, that's not a problem.

jajik: Done. I also added the note as I did in other cases.

jajik:
(I've just rebased to the latest main.)

jfclere (approved):

PR #118, opened 2023-06-30
Add missing slotmem include

jajik:
It's good to remove the removed one, but one should not forget about adding the right one... sorry.


rhusar (approved):
:-D

PR #120, opened 2023-07-07
Rework test without ci

jajik:
This is the first part of work for #112.

PR #121, opened 2023-07-24
Add basic tests into CI

jajik:
This is based on #120 , it adds only the last 4 commits. In those, there is a CI support for running tests and minor tweaks. Once the #120 is merged, I'll change the URL for MPC_CONF to the `main` and then we can merge this too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,3 +139,43 @@ jobs:
       - name: Build
         run: |
           ./native/scripts/windows-build.bat
+  tests:
+    runs-on: ubuntu-latest
+    env:
+      IMG: local-tomcat
+      HTTPD_IMG: local-httpd
+      MPC_CONF: https://raw.githubusercontent.com/jajik/mod_proxy_cluster/rework-test-add-ci/test/httpd/mod_proxy_cluster.conf
+      DEBUG: on
+      ITERATION_COUNT: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup dependencies
+        run: |
+          sudo apt update -y
+          sudo apt install -y moby* maven git curl iproute2
+          cd test
+          sh setup-dependencies.sh
+      - name: Setup images
+        run: |
+          cd test
+          source includes/common.sh
+          httpd_create
+          tomcat_create
+      - name: Basic tests
+        run: |
+          cd test
+          mkdir logs
+          # set vars
+          export MPC_SOURCES=$(git remote -v | grep "origin.*fetch" | awk ' { print $2 } ')
+          export MPC_BRANCH=$(git branch --show-current)
+          # run tests
+          source includes/common.sh
+          run_test basetests.sh "Basic tests"
+      - name: Preserve test logs
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: Test logs
+          path: test/logs/*
+          retention-days: 7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,9 +144,10 @@ jobs:
     env:
       IMG: local-tomcat
       HTTPD_IMG: local-httpd
-      MPC_CONF: https://raw.githubusercontent.com/jajik/mod_proxy_cluster/rework-test-add-ci/test/httpd/mod_proxy_cluster.conf
       DEBUG: on
-      ITERATION_COUNT: 5
+      FOREVER_PAUSE: 30
+      ITERATION_COUNT: 2
+      TOMCAT_CYCLE_COUNT: 2
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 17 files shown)


rhusar:
@jajik So, what's missing here? Can we focus on this PR first so that we can validate the incoming PRs with this having it run in CI? You can ignore (i.e. disable the test) the failure in the broken test for now.

jajik:
@rhusar This is basically ready to go with two notes:

* I wanted to solve #124 at first because the only testcase run in CI is failing because of it. Is is possible to silent this particular job without disabling it?
* I would like to discuss which tests do we want to run. Because currently it's set to `basetests.sh` only and even that takes 11-12 minutes (see https://github.com/jajik/mod_proxy_cluster/actions/runs/5795666167). We might run everything but then we are looking at multiples...

rhusar:
> @rhusar This is basically ready to go with two notes:

Perfect!!

> * I wanted to solve [Tests broken after latest changes #124](https://github.com/modcluster/mod_proxy_cluster/issues/124) at first because the only testcase run in CI is failing because of it. Is is possible to silent this particular job without disabling it?

You can disable the test for now and I will look into that particular test case and fix it (and use GH actions to verify the fix)

I don't think we can silent a job. But it's not that much of a problem, we can force it even with failing tests, but I don't like the failed status on GitHub :)

> * I would like to discuss which tests do we want to run. Because currently it's set to `basetests.sh` only and even that takes 11-12 minutes (see https://github.com/jajik/mod_proxy_cluster/actions/runs/5795666167). We might run everything but then we are looking at multiples...

The WildFly testsuite for instance takes 4-5 hours to run. So it's not really a problem ;-) (until we run out of execution minutes)

rhusar:
> > > @rhusar This is basically ready to go with two notes:
> > 
> > 
> > Perfect!!
> > > * I wanted to solve [Tests broken after latest changes #124](https://github.com/modcluster/mod_proxy_cluster/issues/124) at first because the only testcase run in CI is failing because of it. Is is possible to silent this particular job without disabling it?
> > 
> > 
> > You can disable the test for now and I will look into that particular test case and fix it (and use GH actions to verify the fix)
> > I don't think we can silent a job. But it's not that much of a problem, we can force it even with failing tests, but I don't like the failed status on GitHub :)
> > > * I would like to discuss which tests do we want to run. Because currently it's set to `basetests.sh` only and even that takes 11-12 minutes (see https://github.com/jajik/mod_proxy_cluster/actions/runs/5795666167). We might run everything but then we are looking at multiples...
> > 
> > 
> > The WildFly testsuite for instance takes 4-5 hours to run. So it's not really a problem ;-) (until we run out of execution minutes)
> 
> If this is the case, I'll copy the whole testsuite (test cases) into the CI script and I will disable those that are failing locally. Once we find the issue and I solve #126, I'll simplify the CI job to simply running `testsuite.sh` directly.

Not sure I follow, but what I was getting at is - let's make this PR top priority so that we can test subsequent PRs in CI. In order to get the green passing marks, disable the currently failing test. We can then use the CI to verify the fix (remember, the scope of this PR is getting the tests running in CI and not to fix particular issues).

test/setup-dependencies.sh
```diff
@@ -15,4 +15,4 @@ mvn install || exit 2
 cd $TEST_DIR
 
 # prepare jars
-mvn dependency:copy-dependencies
+mvn install
```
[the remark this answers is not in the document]
jajik: I don't know, it seems like an error.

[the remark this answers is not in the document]
jajik: Nope, it's ok, I changed paths a bit.

jajik:
I changed the workflow so it now uses a matrix. Marking it ready for review, if there is something missing, let me know.

jajik:
I've just added MODCLUSTER-734 test even though it's failing now. That's because of missing load of slotmem module in conf file (fix is part of this PR, so once it's merged, it will pass).

.github/workflows/ci.yml
```diff
@@ -139,3 +139,51 @@ jobs:
...
+      DEBUG: on
+      FOREVER_PAUSE: 30
+      ITERATION_COUNT: 2
+      TOMCAT_CYCLE_COUNT: 2
+    strategy:
+      fail-fast: false
+      matrix:
+        # Those are currently all tests we have. In CI only MODCLUSTER-{734,755}/testit.sh pass, so until we fix the issue, only these tests are enabled.
+        # testcase: [ basetests.sh, hangingtests.sh, maintests.sh, JBCS-1236/testit.sh, MODCLUSTER-640/testit.sh, MODCLUSTER-734/testit.sh, MODCLUSTER-755/testit.sh, MODCLUSTER-785/testit.sh ]
+        testcase: [ MODCLUSTER-734/testit.sh, MODCLUSTER-755/testit.sh ]
```
rhusar, line 156: I see the idea, but that's not going to work here well.... We need a single job with a constant name ('tests' was fine!) so that we can set it as a required check. Moreover, this wastes resources, since the setup is done per test case.

Please just move this in a `for` look above `run_test` and keep this as single execution.

rhusar (changes requested):

.github/workflows/ci.yml
jajik: Ok. I liked it because it was running in parallel thus saving some time (and the setup is not so demanding compared to tests themselves) and results were better accessible.

I updated it as requested.

rhusar (approved):
Looks good to me now

rhusar:
Excellent work, thanks @jajik !

jajik:
@rhusar  Thank you very much for all the help :)

rhusar:
Port merge item - setup 'tests' as required checks - done.

.github/workflows/ci.yml
rhusar: Yeah, the parallelization is nice. We can revisit if we find this an issue, but I don't foresee that.

PR #127, opened 2023-08-08
Drop quay.io

### LINKED ISSUE #125: Use local images instead of quay.io by default
The default should be:

* local images
* no upload to anywhere
* add option to change that (basically we have that, maybe add a function for pushing?)
* update README

jajik:
close #125


rhusar (changes requested):
Looks good but the names are off.

I would drop the "-local" suffix, that's unnecessary. Also, any image can be made not local by pushing it remotely. Also "-image" is unnecessary - the parameter for docker run is an image name. But more importantly, the names are not unique enough - think about what these are for - they are for testing mod_proxy_cluster and they are products of the testsuite. So how about "mod_proxy_cluster-testsuite-tomcat-10.0" and analogous.

test/README.md
```diff
@@ -34,8 +34,8 @@ repository.
 
 Also, if you use `podman` instead of `docker`, make sure you have `podman-docker` package installed (tests are using `docker`).
 
-If you don't want to use quay.io (the default), just set `IMG` and `HTTPD_IMG` variables to docker, local repository or some
-other service.
+By default local `httpd-local` and `tomcat-local` images are used. You can control the names by setting `IMG` and `HTTPD_IMG`
```
rhusar, line 37: Drop the -local.

test/httpd/README.md
```diff
@@ -1,6 +1,6 @@
 # Build the image
 ```bash
-docker build -t quay.io/${USER}/mod_cluster_httpd .
+docker build -t httpd-image .
```
rhusar, line 3: Drop the -image suffice.

test/testsuite.sh
```diff
@@ -12,10 +12,10 @@ if [ -z ${ITERATION_COUNT+x} ]; then
     export ITERATION_COUNT=50 # the number of iteration of starting/stopping a tomcat
 fi
 if [ -z ${IMG+x} ]; then
-    export IMG=quay.io/$USER/tomcat_mod_cluster
+    export IMG=tomcat-local
 fi
 if [ -z ${HTTPD_IMG+x} ]; then
-    export HTTPD_IMG=quay.io/${USER}/httpd_mod_cluster
+    export HTTPD_IMG=httpd-local
```
rhusar, line 18: And here.

jajik:
Point taken. Let's change that if we ever get to that point. :)

rhusar:
> Point taken. Let's change that if we ever get to that point. :)

👍🏼  Agreed.

rhusar:
Merged, thanks @jajik

PR #128, opened 2023-08-09
Fix non-standard log file extension to *.log files.

rhusar:
Doesn't play nice with file extension associations.


jajik (approved):

jajik:
Good catch, thanks.

PR #134, opened 2023-08-10
Testsuite: Make tomcat version configurable, add CI basic test for each version, minor tweaks and fixes

### LINKED ISSUE #131: Testsuite: parametrized tests with supported Tomcat versions
Just the same smoke test (config, autodeploy, status, etc) with different containers would be plenty for now.

jajik:
close #131

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,6 +151,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+# this occupies 8084 port on Ubuntu 20.04 and we don't use it.
+      - name: Remove mono blocking 8084 port
+        run: sudo kill -9 $(sudo lsof -t -i:8084)
       - name: Setup dependencies
         run: |
           sudo apt-get update -y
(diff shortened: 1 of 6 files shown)


rhusar (commented):
So again, great idea, but I am not sure if this is a good fit. In case you want to run this locally, you need to run the testsuite 3x with different parameter. I am under impression that the end goal is running a single command locally to be able to verify with all tests and the same command is then run on CI.

jajik:
Hm, you're right. I intended this as a smoke test into CI, but I guess having an easy option of running this locally would be nice too. I'll change it. Thanks.

jajik:
It's done. I also enabled the whole test suite and made a few fixes.

rhusar:
@jajik This will now need a rebase (final one hopefully!)

jajik:
Around one hour of runtime is expected with the current settings (it can be tweaked using environment variables as we do in `ci.yml`).

jajik:
@rhusar  Rebase is done.

rhusar (approved):
LGTM!

rhusar:
Oh, just to note @jajik these should perhaps use different set of configs, but we don't have to do it right now - https://github.com/modcluster/mod_proxy_cluster/issues/141

jajik:
@rhusar Thanks for the review and merge!

PR #135, opened 2023-08-10
Fix and enable basetests in CI following MODCLUSTER-797 resolution.

rhusar:


rhusar:
Fixed now:

```
Run cd test
Running basetests.sh                               ...  OK
Running MODCLUSTER-734/testit.sh                   ...  OK
Running MODCLUSTER-755/testit.sh                   ...  OK
```

PR #138, opened 2023-08-14
Use local files instead of GH sources

### LINKED ISSUE #126: Testsuite: Use local files in tests, no github references
With that being said, it would be nice if we could run it against the current `main`. That can be handled in `install.sh` ran by the httpd container.

jajik:
close #126


rhusar (approved):
Looks great, thanks @jajik

PR #142, opened 2023-09-19
test: Add option to change tomcat conf files

### LINKED ISSUE #141: Testsuite: all tomcat versions are sharing the same config (and deployment)
This will eventually break, when tomcat adds features that are not supported by previous versions, similar to e.g. jakarta namespace for the deployment.

Ideally, these should have configuration and possibly deployments specific to them. Anyway, so far low priority / optional.

jajik:
close #141. Introduces two variables for server.xml and context.xml by which you can control which files will be used. This does not introduces separate files for each supported version as it is not needed now.

### DIFF
--- a/test/includes/common.sh
+++ b/test/includes/common.sh
@@ -95,8 +95,15 @@ clean_and_exit() {
 #####################################################
 ### T O M C A T   H E L P E R   F U N C T I O N S ###
 #####################################################
+# Function that build a container. IMG variable is used for the container's tag.
+# By passing arguments you can change
+#       $1 tomcat version      (default is 8.5)
+#       $2 tomcat config file  (default is server.xml)
+#       $3 tomcat context file (default is context.xml)
 tomcat_create() {
-    docker build -t $IMG tomcat/ --build-arg TESTSUITE_TOMCAT_VERSION=${1:-8.5}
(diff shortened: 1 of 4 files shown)


jfclere (approved):

rhusar (changes requested):
While this works, I had more of an idea to copy a whole directory overlay (so we can customize anything we want tomcat specific without adding special support for all configuration files we want to override) and have source files in directories like /tomcat-8.5, /tomcat-10.1, etc. WDYT?

jajik:
I don't think it's necessary right now. Are we planning to create different configurations for every tomcat? If not, I think the less files we have to keep track of the better. (And it also prevents config to differ unnecessarily across different versions.)

With that said, I can imagine directories `8.5`, 9.0`, etc. under `test/tomcat/`, that would be clean imo. But it's up to you two, I'll accommodate. :)

jfclere:
@rhusar probably we can do the over stuff in another PR/

rhusar (approved):

rhusar:
> @rhusar probably we can do the over stuff in another PR/

Sure, we can do this later.

rhusar:
Merged, thanks @jajik and @jfclere

PR #143, opened 2023-09-20
Tweak code a bit, remove unnecessary #if macro

jajik:


rhusar:
Thanks @jajik and @jfclere !

PR #145, opened 2023-09-26
Improve tests

jajik:
Just minor tweaks.


test/setup-dependencies.sh
```diff
@@ -2,15 +2,19 @@
...
 git pull --rebase
 mvn install || exit 1
 cp target/websocket-hello-0.0.1.war $TEST_DIR/tomcat/
 cd ..
 
 # get mod_cluster (Java/Tomcat part)
-git clone https://github.com/modcluster/mod_cluster
+if [ ! -d mod_cluster ]; then
+    git clone https://github.com/modcluster/mod_cluster
+fi
```
jfclere, line 17: Should we have warning message if we are using the existing one?

jajik: I don't think so. My intention was to prevent an error when the script is ran again (cloning to a non-empty directory). Using an already present directory/repository is correct imho, the `pull --rebase` will ensure  the content is up-to-date.

jfclere (approved):

jajik:
Thanks @jfclere!

test/setup-dependencies.sh
rhusar: We will probably want to change this altogether an just use maven instead, so whatever here is fine.

PR #146, opened 2023-10-03
Add *.la among cleaned files for make clean

jajik:


jajik:
Closed as this does not solve what I thought it does.

PR #147, opened 2023-10-03
Add mod_lbmethod_cluster.conf.

jfclere:
And add:
Some docs for it.
Arrange DEBUG and some problem when building images.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -75,3 +75,9 @@ There is also a python script that can be run to check mod_proxy_cluster. You ca
 as `miniserver.py`. Execute it simply as `./includes/miniserver.py <port-number>`.
 
 You must have `httpd` with `mod_proxy_cluster` running.
+
+# Testing the lbmethod and mod_proxy_balancer
+Just run the test with CONF=httpd/mod_lbmethod_cluster.conf
+```
+MPC_CONF=httpd/mod_lbmethod_cluster.conf sh testsuite.sh
+```
--- a/test/includes/common.sh
+++ b/test/includes/common.sh
@@ -41,6 +41,16 @@ run_test() {
 httpd_create() {
     rm -rf httpd/mod_proxy_cluster /tmp/mod_proxy_cluster
     mkdir /tmp/mod_proxy_cluster
+    # make sure the native are cleaned before copying
+    for m in advertise mod_proxy_cluster balancers mod_manager
+    do
+        cd ../native/$m
+        echo "Cleaning $m"
+        if [ -f Makefile ]; then
+          make clean
+        fi
+        cd $OLDPWD
+    done
     cp -r ../native ../test /tmp/mod_proxy_cluster/
     mv /tmp/mod_proxy_cluster httpd/
     docker build -t $HTTPD_IMG httpd/
--- a/test/testsuite.sh
+++ b/test/testsuite.sh
@@ -28,6 +28,9 @@ echo "        TOMCAT_CYCLE_COUNT=$TOMCAT_CYCLE_COUNT"
 echo "        ITERATION_COUNT=$ITERATION_COUNT"
 echo "        IMG=$IMG"
 echo "        HTTPD_IMG=$HTTPD_IMG"
+if [ ! -z ${MPC_CONF+x} ]; then
+    echo "        MPC_CONF=$MPC_CONF"
+fi
 
 if [ ! -d logs ]; then
     mkdir logs
@@ -35,8 +38,18 @@ fi
 
 . includes/common.sh
 
-httpd_create  > /dev/null 2>&1 || exit 2
-tomcat_create > /dev/null 2>&1 || exit 3
+if [ ! -d tomcat/target ]; then
+    echo "Missing dependencies. Please run setup-dependencies.sh and then try again"
+    exit 4 
+fi
+
+if [ ! -z ${DEBUG+x} ]; then
+     httpd_create  || exit 2
+     tomcat_create || exit 3
+else
+     httpd_create  > /dev/null 2>&1 || exit 2
+     tomcat_create > /dev/null 2>&1 || exit 3
+fi
 
 # clean everything at first
 echo -n "Cleaning possibly running containers..."
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -1389,8 +1389,7 @@ static char *process_config(request_rec *r, char **ptr, int *errtype)
                 workernode->mess.remove = 0;
                 workernode->mess.num_remove_check = 0;
             } else {
-                loc_unlock_nodes();
-                ap_assert(0); /* we need to figure out what to do... */
+                id = workernode->mess.id;
             }
         } else {
             ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, r->server, "process_config: NEW (%s) %s", nodeinfo.mess.JVMRoute,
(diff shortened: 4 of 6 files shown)


test/includes/common.sh
```diff
@@ -41,6 +41,16 @@ run_test() {
 httpd_create() {
     rm -rf httpd/mod_proxy_cluster /tmp/mod_proxy_cluster
     mkdir /tmp/mod_proxy_cluster
+    # make sure the native are cleaned before copying
+    for m in advertise mod_proxy_cluster balancers mod_manager
+    do
+        cd $m
```
jajik, line 47: There is `native` missing, it should be `cd ../native/$m`.

test/testsuite.sh
```diff
@@ -28,15 +28,28 @@ echo "        TOMCAT_CYCLE_COUNT=$TOMCAT_CYCLE_COUNT"
...
     mkdir logs
 fi
 
 . includes/common.sh
 
-httpd_create  > /dev/null 2>&1 || exit 2
-tomcat_create > /dev/null 2>&1 || exit 3
+if [ -d target ]; then
+  echo "please run setup-dependencies.sh before!!!"
+  exit 1
```
jajik, line 43: `exit 4` would be better; it's better to have different exit code for each situation (see the beginning of the file).

test/testsuite.sh
```diff
@@ -28,15 +28,28 @@ echo "        TOMCAT_CYCLE_COUNT=$TOMCAT_CYCLE_COUNT"
...
 if [ ! -d logs ]; then
     mkdir logs
 fi
 
 . includes/common.sh
 
-httpd_create  > /dev/null 2>&1 || exit 2
-tomcat_create > /dev/null 2>&1 || exit 3
+if [ -d target ]; then
+  echo "please run setup-dependencies.sh before!!!"
```
jajik, line 42: `"Missing dependencies. Please run setup-dependencies.sh and then try again"`, WDYT?

test/testsuite.sh
```diff
@@ -28,15 +28,28 @@ echo "        TOMCAT_CYCLE_COUNT=$TOMCAT_CYCLE_COUNT"
 echo "        ITERATION_COUNT=$ITERATION_COUNT"
 echo "        IMG=$IMG"
 echo "        HTTPD_IMG=$HTTPD_IMG"
+if [ ! -z ${HTTPD_IMG+x} ]; then
```
jajik, line 31: You meant `! -z ${HTTPD_IMG+x}` probably..?

test/README.md
```diff
@@ -75,3 +75,8 @@ There is also a python script that can be run to check mod_proxy_cluster. You ca
 as `miniserver.py`. Execute it simply as `./includes/miniserver.py <port-number>`.
 
 You must have `httpd` with `mod_proxy_cluster` running.
+
+# Testing the lbmethod and mod_proxy_balancer
+Just run the test with CONF=httpd/mod_lbmethod_cluster.conf
+```
```
jajik, line 81: Missing closing ```.

test/includes/common.sh
jfclere: thanks fixed.

test/testsuite.sh
jfclere: fixed.

test/testsuite.sh
jfclere: Oops I meant MPC_CONF

jajik: Yeah, me too :D

test/README.md
jajik: This is still relevant (it displays properly only because its at the end of file)-

jajik (changes requested):
Two things:

* Is the whole testsuite relevant for testing with lbmethod + balancer? If no, I would suggest picking only tests, that are relevant. In both cases, I think it would be good to add test cases directly to the testsuite instead of forcing user to run the testsuite twice.
* It would be nicer to squash commits (there is no need to merge fixed separatedly).

test/README.md
jfclere: https://github.com/jfclere/mod_proxy_cluster/blob/master/test/README.md?plain=1#L83
It is there, what do I miss?

jajik: WTH, I must be blind! I'm sorry, you're right.

jajik:
> > * Is the whole testsuite relevant for testing with lbmethod + balancer?
> 
> Tests that wait for all workers to be removed might be failing but most of the tests should pass.

Ok, that looks like what I'm seeing locally. Let's merge this after applying those two patches above, and I will then incorporate it into the testsuite (so that it is run in CI and the user does not have to run it twice manually).

jfclere:
Actually with mod_lbmethod_cluster.conf there are many:
[Thu Oct 05 09:54:04.191970 2023] [proxy_balancer:error] [pid 12:tid 140571226126016] [client 127.0.0.1:46670] AH10164: balancer://mycluster: failed to add worker ws://127.0.0.1:8080
Probably we need to check if the worker already exits before adding it in this case, no?

jfclere:
I have added the 2 patches I think we are good to merge now ;-)

jajik:
> Actually with mod_lbmethod_cluster.conf there are many: [Thu Oct 05 09:54:04.191970 2023] [proxy_balancer:error] [pid 12:tid 140571226126016] [client 127.0.0.1:46670] AH10164: balancer://mycluster: failed to add worker ws://127.0.0.1:8080 Probably we need to check if the worker already exits before adding it in this case, no?

I briefly looked at it, and it's weird. I'm going to investigate it further once we merge this so we won't delay this PR.

jajik:
> I have added the 2 patches I think we are good to merge now ;-)

I thought you would apply them separately as they are independent of each other. Could you at least change the commit message so it is clear that the assert is removed and logs tweaked? Also, if you removed the "submitted by jajik" part, it would be nice (it's not relevant for the commit message although I truly appreciate it [^1])

[^1]: If you really wish to mention me as a co-author, there is apparently [a way](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) how to do it supported on both, GitHub and Gitlab: just add `Co-authored-by: vchlup@redhat.com` on a separate line in the commit message).

test/testsuite.sh
```diff
@@ -28,15 +28,28 @@ echo "        TOMCAT_CYCLE_COUNT=$TOMCAT_CYCLE_COUNT"
 echo "        ITERATION_COUNT=$ITERATION_COUNT"
 echo "        IMG=$IMG"
 echo "        HTTPD_IMG=$HTTPD_IMG"
+if [ ! -z ${MPC_CONF+x} ]; then
+echo "        MPC_CONF=$MPC_CONF"
```
jajik, line 32: I don't want to be nitpicking, but the echo would be better indented.

test/testsuite.sh
```diff
@@ -28,15 +28,28 @@ echo "        TOMCAT_CYCLE_COUNT=$TOMCAT_CYCLE_COUNT"
...
 
 if [ ! -d logs ]; then
     mkdir logs
 fi
 
 . includes/common.sh
 
-httpd_create  > /dev/null 2>&1 || exit 2
-tomcat_create > /dev/null 2>&1 || exit 3
+if [ -d target ]; then
```
jajik, line 41: Shouldn't there be a `!`?

jajik: Trying it on a fresh repository, it should be `! -d tomcat/target` actually.

jfclere: fixed

test/testsuite.sh
jfclere: fixed.

jajik (approved):
LGTM, thanks for the patience :)

PR #150, opened 2023-10-09
Testsuite tweaks

jajik:
* add lbmethos + mod_proxy_cluster directly into the testsuite (so it is not needed to run tests separately)
 * change naming of the directories for dependencies (so it won't interfere with already checked-out directories for development)
 * fix test fail in case access_log is missing


rhusar (approved):
LGTM

rhusar:
Thanks @jajik

PR #152, opened 2023-10-10
Remove an unnecessary assert

jajik:
It seems that the assert is not needed. The function now returns `NULL` which causes that the outer loop to continue (line 1728 in `internal_find_best_byrequests`).


rhusar (approved):
LGTM

rhusar:
Merged, thanks @jajik

PR #153, opened 2023-10-12
mod_proxy_cluster.c: Code tweaks and polishing, removal of two asserts

jajik:
There are some tweaks, I removed some unnecessary duplication, and removed once-used variables when it made sense. I also removed the two asserts that I think are unnecessary, but it would be better to double check it, I might miss something.

### DIFF
--- a/native/mod_proxy_cluster/mod_proxy_cluster.c
+++ b/native/mod_proxy_cluster/mod_proxy_cluster.c
@@ -302,8 +302,7 @@ static apr_status_t create_worker_reuse(proxy_server_conf *conf, const char *ptr
     helper = *helper_ptr;
     if (helper->index == -1) {
         /* We are going to reuse a removed one */
-        ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, server, "create_worker: reusing removed worker for %s", url);
-        ap_assert(0);
+        ap_log_error(APLOG_MARK, APLOG_NOTICE, 0, server, "create_worker_reuse: reusing removed worker for %s", url);
         return APR_SUCCESS;
     }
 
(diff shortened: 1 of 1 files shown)


native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -186,9 +186,7 @@ static char *normalize_hostname(apr_pool_t *p, const char *hostname)
     return ret;
 }
 
-/**
- * Normalize the worker name
- */
```
jfclere, line 191: I agree the format should be one and we should avoid mixing:
/*
 * Comment
 */
and
/* Comment */

But I would prefer keeping the first one, except you have a good justification.

native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -3156,11 +3090,10 @@ static void upd_context_count(const char *id, int val, const server_rec *s)
     }
 }
 
-static apr_status_t decrement_busy_count(void *worker_)
+static apr_status_t decrement_busy_count(void *w)
```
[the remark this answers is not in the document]
jajik: Ah, ok, so it's a function from the trunk, I did not notice that. Either way, we can leave it like this, we'll erase it in the future anyway.

native/mod_proxy_cluster/mod_proxy_cluster.c
jajik: Looking at it again, I would suggest leaving this as is and creating a new issue to unify the styling for the whole code base (and not only mod_proxy_cluster.c). WDYT?

jfclere: Yes please ;-)

jajik: Ok, I opened #158

native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -1741,7 +1700,7 @@ static proxy_worker *internal_find_best_byrequests(const proxy_balancer *balance
                 qsort(workers, workers_length, sizeof(*workers), &proxy_worker_cmp);
             }
             /* Compute consistent int from session id */
-            for (i = 0; session_id[i] != 0; ++i) {
+            for (i = 0; session_id[i] != 0; i++) {
```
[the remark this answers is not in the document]
jfclere: it is good forget the question...

jfclere (approved):

jajik:
@jfclere Thank you, merged!

PR #154, opened 2023-10-13
mod_manager.c: Refactor part of `process_config`, remove one assert

jajik:
The removal of the assert changes the behaviour; instead of an empty response the server returns `500` now. You can check the behaviour by running httpd with mod_proxy_cluster on port `6666` and executing:

```sh
curl localhost:6666 -H "User-Agent: ClusterListener/1.0" -X CONFIG --data "JVMRoute=demo&Type=http"
curl localhost:6666 -H "User-Agent: ClusterListener/1.0" -X CONFIG --data "JVMRoute=test&ttl=300&Type=http"
```

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -1008,6 +1008,189 @@ static int proxy_node_get_free_id(request_rec *r, int node_table_size)
     return -1;
 }
 
+static void process_config_balancer_defaults(request_rec *r, balancerinfo_t *balancerinfo, mod_manager_config *mconf)
+{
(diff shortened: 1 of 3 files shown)


jajik:
Converting to Draft, there is some issue that needs to be investigated.

jajik:
The issue was caused by an incorrect implementation of a "getter". Hopefully, it's fixed now, but as always thorough check would be great!

jfclere (approved):
Looks OK for me.

jajik:
@jfclere  Thank you very much, merged!

PR #155, opened 2023-10-17
 mod_manager.c: Remove the second assert

jajik:
This PR is based on #154 and it removes the second `ap_assert(0);`. I separated it so it gets enough attention after the first one gets merged.


jajik:
The previous PR was merged, marking as Ready for review.

jfclere (approved):

PR #156, opened 2023-10-19
test: Use the new httpd and Fedora

jajik:
* httpd 2.4.58 is now available, the download link has changed
* use fedora38 image


jajik:
Yes, it was a bit stupid. I dropped the version from the folder, it seems like the best solution. (`WORKDIR` directive is a bit pain and `tar` has a wonderful interface...). If you have a better idea, I'll take it.

rhusar:
> Yes, it was a bit stupid. I dropped the version from the folder, it seems like the best solution. (`WORKDIR` directive is a bit pain and `tar` has a wonderful interface...). If you have a better idea, I'll take it.

That solution looks sufficient to me. Thanks!

rhusar (approved):
LGTM

rhusar:
Thanks @jajik !

PR #162, opened 2023-10-27
Save the error_log and access_log if we are in DEBUG mode.

jfclere:


jajik:
When you run tests through `run_test` function, it will save both of those files in the `logs` directory, isn't that enough?

jfclere:
> When you run tests through `run_test` function, it will save both files in the `logs` directory, isn't that enough?

Yes that is enough.

PR #165, opened 2023-10-31
Fix for issue 163.

jfclere:
when using mod_lbmethod_cluster.c the shared memory for proxy_worker_shared is in mod_balancer, NOT in our shared memory.

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -1206,6 +1206,36 @@ static void mark_node_removed(nodeinfo_t *node)
         node->mess.num_remove_check = 0;
     }
 }
+static const proxy_worker_shared *read_shared_by_node(request_rec *r, nodeinfo_t *node)
+{
+    void *sconf = r->server->module_config;
+    int i, port;
+    char *name = apr_pstrcat(r->pool, "balancer://", node->mess.balancer, NULL);
+    proxy_server_conf *conf = (proxy_server_conf *)ap_get_module_config(sconf, &proxy_module);
+    proxy_balancer *balancer = (proxy_balancer *)conf->balancers->elts;
+    if (sscanf(node->mess.Port, "%u", &port) != 1) {
+        return NULL; /* something is wrong */
+    }
+    for (i = 0; i < conf->balancers->nelts; i++, balancer++) {
+        int j;
+        proxy_worker **workers;
+        if (strcmp(balancer->s->name, name)) {
+            continue;
+        }
+        workers = (proxy_worker **)balancer->workers->elts;
+        for (j = 0; j < balancer->workers->nelts; j++, workers++) {
+            proxy_worker *worker = *workers;
+            ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, r->server, "read_shared_by_node: Balancer %s worker %s, %s, %d",
+                         balancer->s->name, worker->s->route, worker->s->hostname, worker->s->port);
+            if (worker->s->port == port && strcmp(worker->s->hostname, node->mess.Host) == 0 &&
+                strcmp(worker->s->route, node->mess.JVMRoute) == 0) {
+                return worker->s;
+            }
+        }
+    }
+
+    return NULL;
+}
 
 /*
  * Process a CONFIG message
@@ -1406,6 +1436,7 @@ static char *process_config(request_rec *r, char **ptr, int *errtype)
             }
             clean = 0;
             ap_assert(worker->s->port != 0);
+            /* XXX: really needed? offset logic OK here, we save the worker information (see mod_proxy_cluster) */
             pptr = (char *)&nodeinfo;
             offset = sizeof(nodemess_t) + sizeof(apr_time_t) +
                      sizeof(int); /* nodeinfo.offset doesn't contain the information */
@@ -1763,9 +1794,8 @@ static char *process_info(request_rec *r, int *errtype)
 
     for (i = 0; i < size; i++) {
         nodeinfo_t *ou;
-        proxy_worker_shared *proxystat;
+        const proxy_worker_shared *proxystat;
         char *flushpackets;
-        char *pptr;
         if (get_node(nodestatsmem, &ou, id[i]) != APR_SUCCESS) {
             continue;
         }
@@ -1820,9 +1850,11 @@ static char *process_info(request_rec *r, int *errtype)
             break;
         }
 
-        pptr = (char *)ou;
-        pptr = pptr + ou->offset;
-        proxystat = (proxy_worker_shared *)pptr;
+        proxystat = read_shared_by_node(r, ou);
+        if (!proxystat) {
+            ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, r->server, "process_config: No proxystat, assum zeros");
+            proxystat = apr_pcalloc(r->pool, sizeof(proxy_worker_shared));
+        }
 
         switch (type) {
         case TEXT_XML:
@@ -2980,10 +3012,18 @@ static char *process_domain(request_rec *r, char **ptr, int *errtype, const char
 }
 
 /* XXX: move to mod_proxy_cluster as a provider ? */
-static void printproxy_stat(request_rec *r, int reduce_display, const proxy_worker_shared *proxystat)
+static void printproxy_stat(request_rec *r, int reduce_display, nodeinfo_t *node)
 {
-    char *status = NULL;
-    status = proxystat->status & PROXY_WORKER_NOT_USABLE_BITMAP ? "NOTOK" : "OK";
+    char *status;
+    const proxy_worker_shared *proxystat = read_shared_by_node(r, node);
+    if (!proxystat) {
+        ap_log_error(APLOG_MARK, APLOG_ERR, 0, r->server, "printproxy_stat: can't find worker");
+        status = "NOTOK";
+        proxystat = apr_pcalloc(r->pool, sizeof(proxy_worker_shared));
+    } else {
+        status = proxystat->status & PROXY_WORKER_NOT_USABLE_BITMAP ? "NOTOK" : "OK";
+    }
+
 
     if (reduce_display) {
         ap_rprintf(r, " %s ", status);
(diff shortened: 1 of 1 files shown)


jajik:
You can fix the format automatically by running `clang-format -i mod_manager/mod_manager.c`.

native/mod_manager/mod_manager.c
```diff
@@ -1820,9 +1850,11 @@ static char *process_info(request_rec *r, int *errtype)
             break;
         }
 
-        pptr = (char *)ou;
-        pptr = pptr + ou->offset;
```
jajik, line 1824: It's necessary to erase `pptr` declaration on line 1798 too.

native/mod_manager/mod_manager.c
```diff
@@ -1206,6 +1206,35 @@ static void mark_node_removed(nodeinfo_t *node)
...
+    proxy_server_conf *conf = (proxy_server_conf *)ap_get_module_config(sconf, &proxy_module);
+    proxy_balancer *balancer = (proxy_balancer *)conf->balancers->elts;
+    for (i = 0; i < conf->balancers->nelts; i++, balancer++) {
+        int j;
+        proxy_worker **workers;
+        if (strcmp(balancer->s->name, name))
+            continue;
+        workers = (proxy_worker **)balancer->workers->elts;
+        for (j = 0; j < balancer->workers->nelts; j++, workers++) {
+            proxy_worker *worker = *workers;
```
jajik, line 1225: I would consider writing it like this
```c
for (j = 0; j < balancer->workers->nelts; j++) {
    proxy_worker *worker = *(workers + j);
```

I believe it's cleaner.

native/mod_manager/mod_manager.c
```diff
@@ -1206,6 +1206,35 @@ static void mark_node_removed(nodeinfo_t *node)
...
+            ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, r->server, "Balancer %s worker %s, %s, %d",
```
jajik, line 1226: Please add `read_shared_by_node:` at the beginning of the debug message.

native/mod_manager/mod_manager.c
```diff
@@ -1406,6 +1435,7 @@ static char *process_config(request_rec *r, char **ptr, int *errtype)
             }
             clean = 0;
             ap_assert(worker->s->port != 0);
+            /* XXX: needed >?? offset logic OK here, we save the worker information  */
```
jajik, line 1438: I'm not sure I understand the comment. Could you please clarify what do you mean by it?

native/mod_manager/mod_manager.c
```diff
@@ -1820,9 +1850,11 @@ static char *process_info(request_rec *r, int *errtype)
             break;
         }
 
-        pptr = (char *)ou;
-        pptr = pptr + ou->offset;
-        proxystat = (proxy_worker_shared *)pptr;
+        proxystat = read_shared_by_node(r, ou);
+        if (!proxystat) {
+            ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, r->server, "No proxystat, assum zeros");
```
jajik, line 1855: Please, add `process_config:` at the beginning of the debug message.

native/mod_manager/mod_manager.c
```diff
@@ -2980,17 +3012,26 @@ static char *process_domain(request_rec *r, char **ptr, int *errtype, const char
...
 /* XXX: move to mod_proxy_cluster as a provider ? */
-static void printproxy_stat(request_rec *r, int reduce_display, const proxy_worker_shared *proxystat)
+static void printproxy_stat(request_rec *r, int reduce_display, nodeinfo_t *node)
 {
-    char *status = NULL;
-    status = proxystat->status & PROXY_WORKER_NOT_USABLE_BITMAP ? "NOTOK" : "OK";
+    char *status;
+    proxy_worker_shared *proxystat = read_shared_by_node(r, node);
+    if (!proxystat) {
+        ap_log_error(APLOG_MARK, APLOG_ERR, 0, r->server, "printproxy_stat can't find worker");
```
jajik, line 3020: I don't want to be nitpicking, but could you please add `:` after `printproxy_stat` in the debug message?

native/mod_manager/mod_manager.c
```diff
@@ -2980,17 +3012,26 @@ static char *process_domain(request_rec *r, char **ptr, int *errtype, const char
...
-        ap_rprintf(r, ",Status: %s,Elected: %d,Read: %d,Transferred: %d,Connected: %d,Load: %d", status,
-                   (int)proxystat->elected, (int)proxystat->read, (int)proxystat->transferred, (int)proxystat->busy,
-                   proxystat->lbfactor);
+        if (!proxystat)
+            ap_rprintf(r, ",Status: NOTOK,Elected: 0,Read: 0,Transferred: 0,Connected: 0,Load: 0");
+        else
+            ap_rprintf(r, ",Status: %s,Elected: %d,Read: %d,Transferred: %d,Connected: %d,Load: %d", status,
+                       (int)proxystat->elected, (int)proxystat->read, (int)proxystat->transferred, (int)proxystat->busy,
+                       proxystat->lbfactor);
     }
```
jajik, line 3035: This does not have to be nested, `else if (!proxystat) { ... } else { ... }` should be ok.

jajik (changes requested):
Just a few comments.

native/mod_manager/mod_manager.c
jfclere: I think we can remove that code...
https://github.com/modcluster/mod_proxy_cluster/blob/main/native/mod_proxy_cluster/mod_proxy_cluster.c#L858
but the logic in mod_proxy_cluster needs to be checked before....

native/mod_manager/mod_manager.c
jfclere: fixed.

native/mod_manager/mod_manager.c
jfclere: fixed

native/mod_manager/mod_manager.c
jajik: Ok then.

Btw check you chat please :)

native/mod_manager/mod_manager.c
jfclere: mod_balancer/proxy_util something cleaner...
+++
        proxy_worker **worker = (proxy_worker **)balancer->workers->elts;
        for (i = 0; i < balancer->workers->nelts; i++, worker++) {
+++

native/mod_manager/mod_manager.c
jfclere: proxystat = apr_pcalloc(r->pool, sizeof(proxy_worker_shared));
would be better no? we are already testing proxystat to set status.

native/mod_manager/mod_manager.c
jajik: My main "issue" was with the two expressions at the end, but leaving it as is works too, it's not a big deal.

native/mod_manager/mod_manager.c
```diff
@@ -1206,6 +1206,36 @@ static void mark_node_removed(nodeinfo_t *node)
         node->mess.num_remove_check = 0;
     }
 }
+static proxy_worker_shared *read_shared_by_node(request_rec *r, nodeinfo_t *node)
```
jajik, line 1209: Thinking about this returning pointer to the shared memory, I would suggest adding `const`.

native/mod_manager/mod_manager.c
jajik: I don't understand. The `proxystat` is from `read_shared_by_node`, isn't it?

jfclere: yes.

native/mod_manager/mod_manager.c
jfclere: fixed.

native/mod_manager/mod_manager.c
jajik: Ok, now I get it. But this solution has a problem that in case status is NOTOK, then nobody can rely on the rest of the data/information, because it might be 1) valid data, but worker is not usable; 2) the data are "random" because it's just freshly allocated memory. And there is no way how to say which one it is, or am I missing something?

jfclere: it is 1) the worker is NOT yet available in fact.
I first put an assert() here but with mod_lbmethod_cluster/mod_proxy_balancer it happens when starting tomcat...

jfclere: Anyway the status NOTOK is correct ;-) you can't use the worker.

jajik: Ok, but wouldn't it be better to have this instead?

```c
if (reduce_display || strcmp(status, "NOTOK") == 0) {
        ap_rprintf(r, " %s ", status);
} else {
        ap_rprintf(r, ",Status: %s,Elected: %d,Read: %d,Transferred: %d,Connected: %d,Load: %d", status,
                   (int)proxystat->elected, (int)proxystat->read, (int)proxystat->transferred, (int)proxystat->busy,
                   proxystat->lbfactor);
}
```

So that we do not print values that are unusable.

jfclere: well that might break clients parsing the result...

jajik: Ok, that did not occur to me.

native/mod_manager/mod_manager.c
```diff
@@ -1206,6 +1206,36 @@ static void mark_node_removed(nodeinfo_t *node)
...
+static const proxy_worker_shared *read_shared_by_node(request_rec *r, nodeinfo_t *node)
+{
+    void *sconf = r->server->module_config;
+    int i, port;
+    char *name = apr_pstrcat(r->pool, "balancer://", node->mess.balancer, NULL);
+    if (sscanf(node->mess.Port, "%u", &port) != 1) {
+        return NULL; /* something is wrong */
+    }
+    proxy_server_conf *conf = (proxy_server_conf *)ap_get_module_config(sconf, &proxy_module);
+    proxy_balancer *balancer = (proxy_balancer *)conf->balancers->elts;
```
jajik, line 1218: This must be at the beginning of the function in order httpd with maintainer mode to pass.

jajik (changes requested):

native/mod_manager/mod_manager.c
jfclere: fixed.

jajik:
It is necessary to fix styles again and rebase. While rebasing, could you please squash all of those commits into the first one? The arrange commits do not need to be separate.

jajik (approved):
LGTM

PR #171, opened 2023-11-21
Testsuite tweaks: separate websockets, organize directories...

jajik:
* separate websocket test from maintest
* move testapp into shared include folder
* drop "PREFIX functionality"
    * the idea was to permit execution from the main test directory and the individual test directories as well, but it was never used and I don't see any use for it (and it is not in a fully working state anyway)


jfclere (approved):

jajik:
Thanks for the review. Merged.

PR #172, opened 2023-12-05
test: Set httpd container workid to Apache's root

jajik:
It's much more convenient in case you want to log and work within the container.


rhusar (commented):
LGTM

rhusar (approved):

rhusar:
Thanks @jajik

PR #173, opened 2023-12-11
Fix for MODCLUSTER-801

jfclere:
Fix for https://issues.redhat.com/browse/MODCLUSTER-801

mod_proxy_cluster returns 404 now instead of 503 when there's no matching context


rhusar (changes requested):
Needs formatting fixes, see CI results (https://github.com/modcluster/mod_proxy_cluster/actions/runs/7166109210/job/19509427191?pr=173) or run clang locally - https://github.com/modcluster/mod_proxy_cluster?tab=readme-ov-file#styles

PR #177, opened 2023-12-20
Move extra debug macro to the header file, fix logging

jajik:
Two things here

* macro `HAVE_CLUSTER_EX_DEBUG` got moved to the header file where it belongs (that caused formatting to change a bit)
* fix a few warnings when the macro is defined (`int` vs `long` and null balancer)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,18 @@ Tests
 
 The project contains some tests too. You can find them with a separate README in the `test` subdirectory.
 
+Debugging
+---------
+
+For the debugging, you can specify the `LogLevel` within you `httpd.conf` file. In case you want detailed logs,
+set the level to `DEBUG`. In case that would not be enough, `mod_proxy_cluster` uses messages with level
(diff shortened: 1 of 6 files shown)


rhusar (approved):
LGTM, but I ll leave the rest for @jfclere

jfclere (commented):
Won't it be better to remove the HAVE_CLUSTER_EX_DEBUG and use  APLOG_TRACEn (like n = 8)?

jajik:
@jfclere  Ok, so I minimized the usage of the macro with the following:

* DEBUG messages were changed to TRACE4
* ERROR messages were changed to TRACE3

Additionally, there were 2 TRACE8 messages, which I changed to TRACE6 as it seemed more appropriate (https://httpd.apache.org/docs/2.4/mod/core.html#loglevel).

I wanted to remove the macro completely, but we use it in two cases to generate more code than usual (see common.c:412 and mod_manager.c:2855) which would lead to ineffectivity.

jfclere (approved):

jajik:
@jfclere  Thanks for the review, merging.

PR #180, opened 2024-01-18
Record presence/absence of workers within node (#179)

### LINKED ISSUE #179: Record presence/absence of workers within node
We should replace `node_has_workers` with an attribute within node's structure.

jajik:
close #179 

Also, this will require updating apidocs.

### DIFF
--- a/native/include/node.h
+++ b/native/include/node.h
@@ -47,6 +47,7 @@ struct nodemess
     int reversed; /* 1 : reversed... 0 : normal */
     int remove;   /* 1 : removed     0 : normal */
     long ResponseFieldSize;
+    unsigned has_workers;
 
     /* node conf part */
     int flushpackets;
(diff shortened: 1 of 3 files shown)


rhusar:
@jfclere Please review.

jfclere (commented):
Are we sure the node table is looked when using pair_worker_node()/unpair_worker_node()
I would put the if (ou->mess.has_workers > 0) in a function in case we need to use apr atomic at some point.

jajik:
> Are we sure the node table is looked when using pair_worker_node()/unpair_worker_node()?

One call was not called, it should be fixed now.

> I would put the if (ou->mess.has_workers > 0) or any read to has_worker in a function in case we need to use apr atomic at some point.

That's a good point. I reintroduced `node_has_workers` function. The rest of the accesses happens within pair/unpair functions. I left that as is.

jfclere (approved):

jajik:
Merging, thanks for the review.

PR #189, opened 2024-02-01
Use provider for shared version_data

jfclere:
Use provider instead shared memory directly.

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -109,7 +109,7 @@ static const char *node_mutex_type = "node-shm";
 static const char *context_mutex_type = "context-shm";
 
 /* counter for the version (nodes) */
-static apr_shm_t *versionipc_shm = NULL;
+static ap_slotmem_instance_t *version_node_mem = NULL;
 
 /* shared memory */
 static mem_t *contextstatsmem = NULL;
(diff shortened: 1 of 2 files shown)


jajik (commented):
This needs a rebase.

I would consider changing the commit message to actually describe the changes in a few words (if possible), but it's not that important when the issue number is mentioned.

Lastly, after merging we'll need to update apidocs.

I'll review it properly tomorrow.

native/mod_manager/mod_manager.c
```diff
@@ -207,8 +207,25 @@ static apr_status_t loc_find_node(nodeinfo_t **node, const char *route)
...
+    }
+}
+static apr_uint64_t get_version_node(void)
+{
+    version_data *base;
+    if (storage->dptr(version_node_mem, 0, (void **)&base) == APR_SUCCESS) {
+        return base->counter;
+    } else {
+        return 0;
+    }
```
jajik, line 221: We can omit the else branch and leave only `return 0;`

jajik (changes requested):
Looks ok, just a nitpick (and the rebase).

jajik:
I'll check out the behaviour on Windows in the meantime.

jajik:
I'm facing some issues running it on Windows (but this is not limited to your changes, I'm afraid). Could you check it out with #192 please?

jajik:
@jfclere httpd 2.4.59 is officially out (incl. apache lounge), could you please rebase? I've updated the PR mentioned above accordingly.

jajik (approved):
Looks fine. If tests pass, let's merge it. I would only consider changing the commit message for something more descriptive, but that's not a deal breaker. (btw when you prepend `#` before the issue number, GH will link the commit to the issue – see [here](https://github.com/jajik/mod_proxy_cluster/commit/a552acde56459b414a103bd41ed2f90a426f53e5))

PR #190, opened 2024-02-01
Cache context finding

jajik:
I'm opening this as a draft until we discuss the proposal.


jfclere (approved):
Looks good to me.

jajik:
Marking it ready for review. I think we can try merging it. It will give us some performance boost in cases of many contexts and it seems the change is less radical than I though originally.

jajik:
Merging this one as well.

Issue #149, opened 2023-10-06
Testsuite: Investigate failures and make it more stable


jajik:
The testsuite fails occasionally, mainly main tests and JBCS-1236 tests. Investigate and fix those instabilities if possible.

jajik:
It seems that at least some of the failures are linked to tomcat's connector ports. That will be addressed with #140  using `portOffset`.

rhusar:
> It seems that at least some of the failures are linked to tomcat's connector ports (more precisely to the custom mechanism we use). That will be addressed with #140 using `portOffset`.

I will to to get those fixed next week.

jajik:
Ack, thank you.

jajik:
Given the changes we made since opening this issue and the fact that we don't experience random fails anymore, I'm closing the issue now.

PR #191, opened 2024-02-01
Testsuite: Add support for specifying whole addresses for tomcat and proxy

### LINKED ISSUE #188: Testsuite: Allow specifying IP address for tomcat and proxy

jajik:
close #188 

I would also consider closing #149

### DIFF
--- a/test/JBCS-1236/testit.sh
+++ b/test/JBCS-1236/testit.sh
@@ -78,10 +78,10 @@ singlecycle() {
         R=$(expr 1 + $RANDOM % 10 + 10)
         R=$(expr $R + 2)
         # TODO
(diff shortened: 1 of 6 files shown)


jfclere (commented):
What was the problem with 8009? Problem with a running tomcat?

jajik:
> What was the problem with 8009? Problem with a running tomcat?

The issue was that we use `8005` for shutdown. With `portOffset` in use, `tomcat5` would get `8009` for its shutdown port, but that was connector port for `tomcat1`.

jfclere (approved):

jajik:
Merging, thanks @jfclere!

Issue #130, opened 2023-08-10
Testsuite: integrate a proper testing framework rather than using set of custom scripts


rhusar:
e.g. use bats.

jajik:
Just to mention, the current testsuite tries to be POSIX compliant, which bats is not (?). We might consider [ShellSpec](https://github.com/shellspec/shellspec), which should be.

However, I don't know any of these, so I'll have a look and will report about it.

rhusar:
@jajik When you have a minute, if you can look into what upstream Apache and / or it's modules use that would be interesting too. I am going to just assume it's something home brewed and not portable.

jajik:
That's a good idea. Will do.

jajik:
So, it turns out that httpd uses (maybe a bit surprisingly) pytest for some testing (see [README.pytest](https://github.com/apache/httpd/blob/trunk/test/README.pytest)).  That looks interesting, although I don't like python very much. Personally, I don't think rewriting tests we have into pytest is the way to go, I would prefer just adapting them to ShellSpec/bats, but for new tests it might be an option.

rhusar:
> So, it turns out that httpd uses (maybe a bit surprisingly) pytest for some testing (see [README.pytest](https://github.com/apache/httpd/blob/trunk/test/README.pytest)). That looks interesting, although I don't like python very much. Personally, I don't think rewriting tests we have into pytest is the way to go, I would prefer just adapting them to ShellSpec/bats, but for new tests it might be an option.

If we are talking about preferences I would strongly to prefer to have them in Java. I am under impression that what we have now can be done in Java quite with ease.

jfclere:
https://github.com/apache/httpd-tests is the one I contributed.

jajik:
That looks like Perl! Definitely prefer that over Python, I'll have a look, thanks!

PR #192, opened 2024-02-06
CI: Run httpd with mod_proxy_cluster even on Windows

jajik:
We only build the project on Windows. That means we might (and we did)  miss errors that would be detected when we would run httpd with mod_proxy_cluster. This PR extends the Windows job and after building mod_proxy_cluster runs it trying two request through `curl`.

I open this as a draft, because it's failing currently (and the error is at our side, without loading mod_proxy_cluster and the second `curl` it works).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,7 +177,7 @@ jobs:
     runs-on: windows-latest
     env:
       APACHE_LOUNGE_DISTRO_VERSION: 2.4.57
-      HTTPD_DEV_HOME: '${{ github.workspace }}\httpd-apache-lounge\Apache24'
+      HTTPD_DEV_HOME: 'C:\Apache24'
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


jajik:
I have no idea how common are use cases with other platforms involved (for my sanity, I hope especially Windows are seriously limited).

However, I don't think running it on other platforms would bring any benefits because we're using containers anyway. For #130 it might be relevant to focus on native tests as well.

rhusar:
Just to check, the last commit 'Update httpd to 2.4.59 on Windows' is what fixed the issue that we have been hitting before, correct?

jajik:
> Just to check, the last commit 'Update httpd to 2.4.59 on Windows' is what fixed the issue that we have been hitting before, correct?

That and also #189

jfclere (approved):

rhusar (approved):
Excellent work @jajik

rhusar:
Thanks @jajik and @jfclere - merging this first so that we can run the test before merging subsequent changes.

jajik:
Thanks! I was just about to merge #202 :smile:

rhusar:
> Thanks! I was just about to merge #202 😄

Right, better to merge a working test first - if a subsequent change breaks the test, the onus is then on that person, rather than on the one adding the test.

PR #199, opened 2024-02-08
Fix Conditional jump or move depends on uninitialised value(s) warning

jajik:
The warning comes from valgrind and is imho relevant (although the functionality does not change). There is no point in going beyond `'\0'` if present.


rhusar (approved):
LGTM - btw is this the only place where this is done in this manner?

jajik:
Yes.

@jfclere any objections?

jfclere (approved):

jajik:
Thanks, merging.

PR #200, opened 2024-02-13
Tweak watchdog logic, drop set_interval, free allocated memory

jajik:
A few notes:

* the deallocation should happen only if the check is run on a thread, otherwise the `thread` pointer is `NULL` and the `if` won't execute
* the `default` added should not be necessary, but it won't hurt if something gets redefined (AP_STATEs are only macros...)
* I don't think that set_interval is something we need, please double check me on that


jfclere (approved):
It is perfect now!

jajik:
Thank you for the review, I'll merge it once tests finish.

PR #205, opened 2024-02-21
Drop ap_proxygetline, ap_proxy_make_fake_req, and use buckets directly

### LINKED ISSUE #201: Investigate whether we can drop `ap_proxygetline`
so that we can drop code duplication (copy-paste).

https://github.com/modcluster/mod_proxy_cluster/blob/7042993c46605db25142e0bb270231a4ca37cf46/native/mod_proxy_cluster/mod_proxy_cluster.c#L956

jajik:
Functions `ap_proxygetline` and `ap_proxy_make_fake_req` that were copied from mod_proxy were removed, because they are no longer needed. They were replaced with explicit bucket API calls. I tried to clear the code a bit.

close #201

### DIFF
--- a/native/mod_proxy_cluster/mod_proxy_cluster.c
+++ b/native/mod_proxy_cluster/mod_proxy_cluster.c
@@ -982,115 +982,95 @@ cleanup:
     return status;
 }
 
-static apr_status_t ap_proxygetline(apr_bucket_brigade *bb, char *s, int n, request_rec *r, int fold, int *writen)
+static void copy_and_trim_crlf(char *buffer, const char *ptr, apr_size_t len)
 {
-    char *tmp_s = s;
-    apr_status_t rv;
-    apr_size_t len;
-
-    rv = ap_rgetline(&tmp_s, n, &len, r, fold, bb);
-    apr_brigade_cleanup(bb);
-
-    if (rv == APR_SUCCESS) {
-        *writen = (int)len;
-    } else if (rv == APR_ENOSPC) {
-        *writen = n;
-    } else {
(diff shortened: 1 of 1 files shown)


jajik:
@jfclere There is a little difference from the previous solution as the received log message (line 1041) is now logged as a whole (even though `AP_MODE_GETLINE` is used), and I'm not sure why. I'm going to dig into it again a little bit more tomorrow.

native/mod_proxy_cluster/mod_proxy_cluster.c
```diff
@@ -982,115 +982,66 @@ static apr_status_t ajp_handle_cping_cpong(apr_socket_t *sock, const request_rec
...
-    rp = ap_proxy_make_fake_req(p_conn->connection, r);
-    rp->proxyreq = PROXYREQ_RESPONSE;
-    tmp_bb = apr_brigade_create(r->pool, p_conn->connection->bucket_alloc);
-    while (1) {
-        ap_proxygetline(tmp_bb, buffer, sizeof(buffer), rp, 0, &len);
-        if (len <= 0) {
+    while (!APR_BRIGADE_EMPTY(brigade)) {
+        e = APR_BRIGADE_FIRST(brigade);
+        if (APR_BUCKET_IS_EOS(e)) {
+            ap_log_error(APLOG_MARK, APLOG_DEBUG, rv, r->server, "http_cping_cpong: bucket is EOS!");
```
jajik, line 1031: This should be dropped.

jfclere (approved):

jajik:
Thank you for the review

PR #207, opened 2024-02-27
Minor typo fix in LBstatusRecalTime option description.

rhusar:
Term "load balancing" is two separate words.


jajik (approved):
Nice catch

rhusar:
Thanks @jajik

PR #210, opened 2024-03-06
test: Fix wrong path and unnecessary sed for tomcat container

jajik:
The conPort is no longer used. And `bin/catalina.sh` is the right part (see the conf above or WORKDIR in the dockerfile).


rhusar (approved):
LGTM

rhusar:
Thanks @jajik

PR #213, opened 2024-03-13
Refactor updating worker's lbstatus

jajik:
This PR makes the main `update_workers_lbstatus` function a bit wordier, however, it might be more clear now with the check logic separated. My main motivation was to make (un)locking a bit cleaner.


jfclere (approved):

jajik:
Thanks for the review.

PR #216, opened 2024-03-26
Tweak LogLevels to prevent logfile flooding

jajik:
I've found another cases where we log more than necessary. These are special because they can (in some cases) occur frequently and are using notice level which is unfortunate as they are logged even higher levels are specified.

> When logging to a regular file, messages of the level notice cannot be suppressed and thus are always logged. However, this doesn't apply when logging is done using syslog.

from https://httpd.apache.org/docs/2.4/mod/core.html#loglevel


jajik:
There are a few notice levels remaining, but these seems reasonable (and shouldn't happen too often, so I think they can stay as they are).

jfclere (approved):

jajik:
Thanks for the review! Merging.

PR #219, opened 2024-04-02
Use AP_DECLARE_MODULE macro

jajik:
It is equivalent to `AP_MODULE_DECLARE_DATA` used currently with the addition of setting the name of the module for the logs (see [the definition](https://github.com/apache/httpd/blob/21155482135ad2ba8e4bfc547dd6d6a0093d502f/include/http_config.h#L467)).

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -4030,7 +4030,7 @@ static void *merge_manager_server_config(apr_pool_t *p, void *server1_conf, void
 
 /* Module declaration */
 
-module AP_MODULE_DECLARE_DATA manager_module = {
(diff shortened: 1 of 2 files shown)


jajik:
So for example we'll have in logs this

```
[Tue Apr 02 11:15:12.840811 2024] [manager:warn] [pid 15:tid 140020407076544] process_appl_cmd: ENABLE: context /legacy is in balancer  and mycluster
```

instead of this

```
[Tue Apr 02 11:15:12.840811 2024] [:warn] [pid 15:tid 140020407076544] process_appl_cmd: ENABLE: context /legacy is in balancer  and mycluster
```

jfclere (approved):

jajik:
Thanks, merging.

PR #225, opened 2024-04-19
test: Add a dependecy check for used commands

jajik:
When I tried running the testsuite on a fresh system (ubuntu), I forgot to install `curl`. The testuite then failed because it assumed httpd was not ready, but in reality it was missing `curl`. This PR adds a check for the required commands into the dependencies setup.

### DIFF
--- a/test/setup-dependencies.sh
+++ b/test/setup-dependencies.sh
@@ -1,3 +1,26 @@
+#!/usr/bin/sh
+
+# $1 is the command, $2 is the flag used for the check (--version by default)
+check_cmd() {
+if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


rhusar (approved):
Why not, although, this is documented

jajik:
Well, there exists people like me who even though open the docs, they forget to install some of the tools... :smile:

Thanks for the merge.

rhusar:
> Well, there exists people like me who even though open the docs, they forget to install some of the tools... 😄
> 
> Thanks for the merge.

Yeah my take is that it saves a lot of unnecessary debugging in those cases

PR #228, opened 2024-04-29
ci: upgrade to Fedora 40; remove unnecessary 'podman pull' to avoid v…

rhusar:
…ersion declaration duplication.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,11 +23,10 @@ jobs:
         run: |
           sudo apt update
           sudo apt-get -y install podman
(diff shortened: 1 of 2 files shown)


jajik (approved):
We might use simply the latest, but this is fine.

rhusar:
> We might use simply the latest, but this is fine.

That's intentional to use an explicit version to have as close to reproducible environment as possible. Even specifying a version here is not a guarantee of stability, but if Fedora keeps everything compatible, there shouldn't be any breakage.

rhusar:
Thanks @jajik

PR #229, opened 2024-05-21
ci: Use ubuntu-24.04 instead of macOS for clang-format

### LINKED ISSUE #222: CI: Drop macos runner for clang-format
Once Ubuntu 24.02 will come out, it should have clang-format new enough to drop the macOS runner.

jajik:
close #222 

The runner for the newest Ubuntu is available (it's Beta officially, but for clang-format it's not a big deal), so let's move away from macOS.


rhusar (approved):
Agreed!

rhusar:
Thanks @jajik !

PR #231, opened 2024-05-29
ci: Add -Werror and -Wall to trunk and maintainer mode

### LINKED ISSUE #230: Introduce -Werror flag to CI

jajik:
close #230 

It wasn't as straightforward as expected. I did not add it into the cmake option, because for gcc it is the same, and for clang it will fail, because it complains about some apr stuff.


rhusar:
> it complains about some apr stuff (which is outside of our scope).

It sounds like the situation example I was talking about on Tuesday... nevertheless, this is in CI which we are free to change and update at any time.

rhusar (approved):

rhusar:
Thanks @jajik !

FYI @jfclere

PR #234, opened 2024-06-04
MODCLUSTER-204: Increase supported Alias length and add warning if exceeded

jajik:

PR #236, opened 2024-06-07
Follow up on MODCLUSTER-204

jajik:
After MODCLUSTER-204 fix in #234, I've noticed that there's no warning in case of too long context. This adds one & fixes a small off-by-one in printing.

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -740,7 +740,7 @@ static apr_status_t insert_update_host_guard(server_rec *s, mem_t *mem, hostinfo
     int len = strlen(alias);
     if (len > HOSTALIASZ) {
         ap_log_error(APLOG_MARK, APLOG_WARNING, 0, s,
-                     "process_config: received alias %s is too long (trimmed to 255 characters)", alias);
+                     "process_config: received alias %s is too long (trimmed to %d characters)", alias, HOSTALIASZ);
     }
     strncpy(info->host, alias, HOSTALIASZ);
     info->host[HOSTALIASZ] = '\0';
@@ -809,7 +809,7 @@ static apr_status_t insert_update_context_guard(server_rec *s, mem_t *mem, conte
 
     if (len > CONTEXTSZ) {
         ap_log_error(APLOG_MARK, APLOG_WARNING, 0, s,
-                     "process_config: received context %s is too long (trimmed to 80 characters)", context);
+                     "process_config: received context %s is too long (trimmed to %d characters)", context, CONTEXTSZ);
     }
 
     return insert_update_context(mem, info);
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -794,53 +794,64 @@ static void read_remove_context(mem_t *mem, contextinfo_t *context)
     }
 }
 
+static apr_status_t insert_update_context_guard(server_rec *s, mem_t *mem, contextinfo_t *info, char *context,
+                                                int status)
+{
+    int len = strlen(context);
+
+    info->id = 0;
+    strncpy(info->context, context, CONTEXTSZ);
+    info->context[CONTEXTSZ] = '\0';
+    if (status == REMOVE) {
+        read_remove_context(mem, info);
+        return APR_SUCCESS;
+    }
+
+    if (len > CONTEXTSZ) {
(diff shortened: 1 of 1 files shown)


native/mod_manager/mod_manager.c
```diff
@@ -794,53 +794,63 @@ static void read_remove_context(mem_t *mem, contextinfo_t *context)
...
 }
 
+static apr_status_t insert_update_context_guard(server_rec *s, mem_t *mem, contextinfo_t *info, char *context,
+                                                int status)
+{
+    int len = strlen(context);
+    if (len > CONTEXTSZ) {
+        ap_log_error(APLOG_MARK, APLOG_WARNING, 0, s,
+                     "process_config: received context %s is too long (trimmed to 255 characters)", context);
+    }
```
jajik, line 804: What about moving it after the remove return? Does it make sense to not inform user about removing too long contexts? (they would get it during insert)

jfclere (approved):
Don't forget to write tests for those ;-)

native/mod_manager/mod_manager.c
jfclere: If you go that way I suggest returning an error to tomcat and add the corresponding error message to the protocol.
But that looks OK for the moment.

jajik: @rhusar JFC is ok with the current state, what do you think?

rhusar: I need to refresh my memory on how this works. So what's the point of trimming the context? If we only have a map of trimmed contexts, we would not be able to make a correct load balancing decision. So that should qualify as an error and thus perhaps fail the operation. 

What was the behavior before these https://github.com/modcluster/mod_proxy_cluster/pull/234 changes went in? Did we just fail or trim?

rhusar (changes requested):

native/mod_manager/mod_manager.c
jajik: The behaviour from the native part was to trim (fixed size memory) without saying anything.

jajik: Now it throws a warning once it is added. The question is whether it makes sense to warn again upon removing or not.

rhusar: > The behaviour from the native part was to trim (fixed size memory) without saying anything.

OK. So not issuing a warning on adding is clearly wrong - so kudos for finding this and the fix. I don't think we need to warn on removing. 

But I am wondering if this should not fail instead.

jajik: Ok, so I'll move the warning after the return and merge it.

We can discuss the (change of) behavior today.

rhusar: > I don't think we need to warn on removing.

More I think about it, the removing might be actually even worse. Anyway, let's discuss on the call later.

jajik: Well, it might end up with removing something else (same prefix of length >= CONTEXTSZ) . But then there would be conflict even earlier, wouldn't be?

> Anyway, let's discuss on the call later.

+1

rhusar: Exactly. The question is what happens when multiple contexts are trimmed to the same string. Are they duplicated in the registry, error or only 1 remains?

jajik: It depends on the alias sent along the context. If it is the same, nothing is added and it is treated as the already existing one. If the context is different, new virtual host is added with the same context but different alias. There's never an error, so this PR adds at least a warning.

jajik:
The last commit improves the warning messages so that they don't get outdated in case we change the macro values again.

native/mod_manager/mod_manager.c
jfclere: https://github.com/modcluster/mod_proxy_cluster/blob/main/native/common/common.c#L443

So basically the context WILL never be routed by mod_proxy_balancer.

rhusar (approved):
OK I am going to approve this since we have a Jira that we need to get to the bottom of - https://issues.redhat.com/browse/MODCLUSTER-822

native/mod_manager/mod_manager.c
rhusar: I ll resolve this and continue on https://issues.redhat.com/browse/MODCLUSTER-822

jajik:
Thanks, merging.

PR #246, opened 2024-07-02
Add RELEASING.md with necessary instructions

### LINKED ISSUE #244: Add RELEASING.md file with step-by-step instructions
It should also include information about versioning from https://developer.jboss.org/docs/DOC-10725

jajik:
close #244

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,31 @@
+# Releasing new version
+
+## Step-by-step
+
+In order to release a new version, there are several steps that must happen:
+
+ 0. make sure all issues (in GitHub and in JIRA) targeted to the version are resolved
+ 1. create a pull request in which you change the [version string](https://github.com/modcluster/mod_proxy_cluster/blob/main/native/include/mod_proxy_cluster.h#L21) according to the format described in [another section](#version-naming-conventions)
+ 2. get review and merge the pull request
+ 3. release the version through GitHub release functionality with a new tag corresponding to the version string from
+ the pull request (create the tag during the release through the release dialog)
+ 4. mark the corresponding version[^1] released in [JIRA](https://issues.redhat.com/projects/MODCLUSTER?release-page) as well
+
+
+## Version naming conventions
+
+Version string must have the following format:
+
+* `major.minor.micro.Dev`
+* `major.minor.micro.Alpha<number>`
+* `major.minor.micro.Beta<number>`
+* `major.minor.micro.CR<number>`
+* `major.minor.micro.Final`
+
+where `<number>` is a mandatory numerical value starting from `1`. Any version marked `Final` must
+not have a number. If there is any issue with such a version, just increase a micro version and go
+through the process again.
+
+Versions in between of `x.y.z.Final` and next `a.b.c.Alpha1` will be called `a.b.c.Dev`.
+
+[^1]: The corresponding version within the JIRA project has `native-` prefix. For example version `2.0.0.Final` in this repository will be `native-2.0.0.Final` in the JIRA project.


jfclere (approved):

RELEASING.md
```diff
@@ -0,0 +1,24 @@
+# Releasing new version
+
+## Step-by-step
+
+In order to release a new version, there are several steps that must happen:
+
+ 0. make sure all issues (in GitHub and in JIRA) targeted to the version are resolved
+ 1. create a pull request in which you change the [version string](https://github.com/modcluster/mod_proxy_cluster/blob/main/native/include/mod_proxy_cluster.h#L21) according to the format described in [another section](#version-naming-conventions)
```
rhusar, line 8: We need to define here what will the snapshot versions be called.

RELEASING.md
```diff
@@ -0,0 +1,24 @@
...
+ 2. get review and merge the pull request
+ 3. release the version through GH release functionality with the tag corresponding to version from the pull request
```
rhusar, line 10: The "GH release functionality" doesn't tag. Just creates a page for the tag.

We are missing the `git tag` + push step.

RELEASING.md
```diff
@@ -0,0 +1,24 @@
...
+ 4. mark the corresponding version released in [JIRA](https://issues.redhat.com/projects/MODCLUSTER?release-page) as well
```
rhusar, line 11: We need to define what the names will look like in Jira (we are currently using prefix native-1.2.3.Final)

rhusar (changes requested):

RELEASING.md
jajik: I think we're good using it in the future as well, but I'll add a note. Good catch.

RELEASING.md
jajik: I know, I'd like to use the hash here. I'll see what can we do here.

RELEASING.md
jajik: You can create a tag directly through it.

jajik: See here https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository?tool=webui (it's not the most intuitive, but it works).

rhusar: OK I see what you are referring to:

> To create a new tag, type a version number for your release, then click Create new tag.

Well, lets test this first and if that works, les document this in more detail.

jajik: I used it before, it works. But I can add a note as well.

rhusar: Right, I believe them it works, but I want to verify whether and how it's signing the tag.

jajik: Are we signing tags? It looks similar to this, if the commit is signed, it shows a green tick https://github.com/modcluster/mod_cluster/releases/tag/2.0.4.Final

jajik:
@rhusar Looking at expected checks, we need to change the settings after #248

rhusar:
Thats funny, doesn't seem to be very well thought out. The only option is to remove all required checks which in turn disables other functionality..

jajik:
What? :D

jajik:
So it seems the workaround is this

https://github.com/laurencee/GithubActionTesting/commit/876ef30bc6aff6f981ff829dbee72f3ef2e2714c

Seriously :smiley:

rhusar:
> Thats funny, doesn't seem to be very well thought out. The only option is to remove all required checks which in turn disables other functionality..

Just for record, it disables this one:

    Require branches to be up to date before merging
    This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).

jajik:
> Just for record, it disables this one:
> 
> ```
> Require branches to be up to date before merging
> This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).
> ```

So you cannot enforce linear history without that?

rhusar:
> > Just for record, it disables this one:
> > ```
> > Require branches to be up to date before merging
> > This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).
> > ```
> 
> So you cannot enforce linear history without that?

I think you mean by 'linear history' changes being tested linearly / on top of each other and IIRC yes, this would remove the notification about branches not being up 2 date, which given we don't have a large turnover in this repo we want to keep. 

I am thinking now let's go with your workaround.

rhusar:
> > > Just for record, it disables this one:
> > > ```
> > > Require branches to be up to date before merging
> > > This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).
> > > ```
> > 
> > 
> > So you cannot enforce linear history without that?
> 
> I think you mean by 'linear history' changes being tested linearly / on top of each other and IIRC yes, this would remove the notification about branches not being up 2 date, which given we don't have a large turnover in this repo we want to keep.
> 
> I am thinking now let's go with your workaround.

OK merged - please rebase and see if this works.

rhusar:
> > > > Just for record, it disables this one:
> > > > ```
> > > > Require branches to be up to date before merging
> > > > This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).
> > > > ```
> > > 
> > > 
> > > So you cannot enforce linear history without that?
> > 
> > 
> > I think you mean by 'linear history' changes being tested linearly / on top of each other and IIRC yes, this would remove the notification about branches not being up 2 date, which given we don't have a large turnover in this repo we want to keep.
> > I am thinking now let's go with your workaround.
> 
> OK merged - please rebase and see if this works.

OK thanks, the workaround looks good - and it's clear that the checks were indeed skipped.

RELEASING.md
rhusar: Adding dynamic git commit hash is really aspect of the build scripts, that infer the git head hash and use it during compilation in the version somewhere.

All we need to do here within the context of this document, is to figure out, what we put in the version string in between actual tags.

rhusar (changes requested):

RELEASING.md
rhusar: > Are we signing tags? It looks similar to this, if the commit is signed, it shows a green tick https://github.com/modcluster/mod_cluster/releases/tag/2.0.4.Final

Doing a quick check - actually we aren't signing tags - because our maven plugin configuration doesn't sign them on tagging although the feature is available since 2018 (https://issues.apache.org/jira/browse/SCM-486)

RELEASING.md
jajik: > Adding dynamic git commit hash is really aspect of the build scripts, that infer the git head hash and use it during compilation in the version somewhere.
>
> All we need to do here within the context of this document, is to figure out, what we put in the version string in between actual tags.

That would need git, which we don't have to have. What I was thinking about was something like pre/commit script that would do it automatically. But that would be noisy, imho (maybe post PR? I haven't explored it yet).

jajik: Maybe we might keep it simple and use just something like `major.minor.micro.Dev` with an optional number at the end. We would increment it once in a while if we make some major changes...

rhusar: > > Adding dynamic git commit hash is really aspect of the build scripts, that infer the git head hash and use it during compilation in the version somewhere.
> > All we need to do here within the context of this document, is to figure out, what we put in the version string in between actual tags.
> 
> That would need git, which we don't have to have. What I was thinking about was something like pre/commit script that would do it automatically. But that would be noisy, imho (maybe post PR? I haven't explored it yet).

Yeah, that can be done with GH Action too, but way too noisy and hacky.

rhusar: > Maybe we might keep it simple and use just something like `major.minor.micro.Dev` with an optional number at the end. We would increment it once in a while if we make some major changes...
> 
> For example after we release `2.0.0.Final`, the next version string would be `2.0.1.Dev`. If we make only small changes, we can do `2.0.1.Dev1`, `2.0.1.Dev2`.... and if we make some bigger ones, we can do `2.1.0.Dev`. Or we can have only Dev and increment only the micro if we don't insist on strict continuity.

I guess it doesn't really matter,  I am mostly concerned about protocol negotiation, but that will be handled by parsing the 'MCMP protocol version' (which is currently reported as `0.2.1`). If the version string itself is to be parsed out, we really care for the Final releases. So, anything is fine really.

jajik: In httpd they have `STATUS` file and a header macros. You can see here how does release of 2.4.61 look like

https://github.com/apache/httpd/commit/d49b8e01d885f85bdcec355190fbcf066d0afc50

version got increased to 2.4.62 with "In Development" status.

jajik: @jfclere @rhusar I've updated the RELEASING.md following the discussion.

rhusar: 👍🏼

PR #247, opened 2024-07-03
ci: Tweak ci so it doesn't run without meaningful changes

jajik:
It occurred to me that we run CI on all commits regardless the changes. This PR will cause CI running only if anything within the CI file, or `native/`, `test/` directories changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,19 @@ on:
   push:
     branches:
       - '*'
+    paths:
+      - test/**
+      - native/**
+      - .github/workflows/ci.yml
+      - Doxyfile
   pull_request:
     branches:
       - '*'
+    paths:
+      - test/**
+      - native/**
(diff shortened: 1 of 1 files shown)


rhusar (changes requested):
So no verifying building docs when Doxyfile is changed? :)

jajik:
> So no verifying building docs when Doxyfile is changed? :)

Fixed.

rhusar (changes requested):
Still not a great approach. Going forward, we would need to keep in mind that adding new paths need to be added for CI scripts.

Rather than selecting which files to run the build for, we can just label the ones that definitely do not effect the build.

Opened https://github.com/modcluster/mod_proxy_cluster/pull/248

rhusar:
For example, this would still run tests for `tests/README.md`. Closing in favor of https://github.com/modcluster/mod_proxy_cluster/pull/248

PR #248, opened 2024-07-03
CI: skip checks on files which do not effect the build nor tests such…

rhusar:
… as markdown files

Replaces https://github.com/modcluster/mod_proxy_cluster/pull/247

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,17 @@ on:
   push:
     branches:
       - '*'
+    paths-ignore:
+      - '.gitignore'
+      - '**.md'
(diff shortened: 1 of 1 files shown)


jajik (commented):
What about `.gitattributes` and `.github/dependabot.yml`?

rhusar:
> What about `.gitattributes` and `.github/dependabot.yml`?

The .gitattributes has direct effect what the checkout looks like and effects the build. 

The dependabot.yml changes are being checked for PRs and I am not sure if ignoring it here wouldn't break those checks.

jajik:
Ok

jajik:
Merging now. Thanks.

PR #250, opened 2024-07-03
Upgrade to httpd 2.4.61

jajik:
close #245

PR #251, opened 2024-07-04
Ci workaround ignored paths

jajik:
followup for #248 and #246

### DIFF
--- a/.github/workflows/ci-ignored-workaround.yml
+++ b/.github/workflows/ci-ignored-workaround.yml
@@ -1,0 +1,45 @@
+name: Skip CI on ignored paths
+on:
+  push:
+    branches:
+      - '*'
+    paths:
+      - '.gitignore'
+      - '**.md'
+      - 'LICENSE'
+  pull_request:
+    branches:
+      - '*'
+    paths:
+      - '.gitignore'
+      - '**.md'
+      - 'LICENSE'
+
+jobs:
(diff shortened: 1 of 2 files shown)


jajik:
The pipeline will be failing until we merge #250

rhusar (changes requested):
LOL, thanks for the hack demonstration, this would work. But oh man, 50 LOC hack to manage just to solve a non-existent problem is not great use of our time. 

Let me try couple of other things, if not, I ll revert the original change.

rhusar (approved):
Lets try this until GH fixes this (even though it's been long time in the making).

rhusar:
BTW I did toy around with the idea of creating a first step to get list of all changes files, if they match any of these, then skip the following jobs. But that's more error prone than this approach, so I scratched that.

jajik:
> Lets try this until GH fixes this (even though it's been long time in the making).

You mean like this one https://github.com/orgs/community/discussions/12032 ? :smiley:

jajik:
Thanks for the merge

rhusar:
> > Lets try this until GH fixes this (even though it's been long time in the making).
> 
> You mean like this one https://github.com/orgs/community/discussions/12032 ? 😃

The rebase button is crap - that invalidates signing information. Using a merge commit in turn adds a lot of noise with merge commits. Sigh..

PR #260, opened 2024-08-02
Remove a double unlock

jajik:
There is another unlock five line below which is the correct one.


jfclere (approved):
Good catch

jajik:
Thanks. I'm afraid there might be more, but I'll create a separate PR in that case.

PR #261, opened 2024-08-05
Improve nodes (un)locking

jajik:
Opening as a draft, because there might be more.


jajik:
> @jajik Unrelated - but can you look into replacing strings such as "text/plain" with macros as well? I see you are changing lines that include it and it's repeated many times.. Thx.

Proposed in #266

PR #262, opened 2024-08-07
Fix handling of marking nodes removed

jajik:
There is `removed = id` statement but `0` is a valid `id`. Therefore for such `id` the code in `if` wouldn't execute.


jfclere (approved):

jajik:
Thanks for the review!

PR #263, opened 2024-08-12
Use defined macro instead of magic contants

jajik:


rhusar (commented):
LGTM, but I defer to @jfclere

jfclere (approved):

PR #264, opened 2024-08-12
test: Increase file limit for httpd container

jajik:
Minor tweak that prevents too many open files error (it doesn't affect our tests, but when you try to stress mod_proxy_cluster using our container, it may have an effect).


rhusar (approved):
I guess I would have added a comment, but that's fine.

rhusar:
Thanks @jajik

PR #267, opened 2024-08-15
Clear the worker status when reusing cleaned worker

jajik:
The current behaviour is suspicious. The call to `ap_proxy_initialize_worker` has basically no effect and without this change the connection to backend fails from time to time (when running perf tests with dynamic worker removal) which causes `all workers in error state` (= 503) even though a corresponding node to the worker is up and running.


jajik:
I'm missing something here, converting to draft.

jajik:
So it seems that the missing part was #269. Ready for review now.

jfclere (approved):

jajik:
Thanks for the reviews!

PR #269, opened 2024-08-16
Fix condition in find_node_context_host

jajik:
I believe the current condition is incorrect, because in case of `balancer-s->name` is shorter than the prefix, it gets included for `node->mess.balancer` even though they don't match.

In combination with #267 it prevents 404 when the node is up but without app (yet).

### DIFF
--- a/native/common/common.c
+++ b/native/common/common.c
@@ -434,7 +434,7 @@ node_context *find_node_context_host(request_rec *r, const proxy_balancer *balan
             if (node == NULL) {
                 continue;
             }
-            if (strlen(balancer->s->name) > BALANCER_PREFIX_LENGTH &&
+            if (strlen(balancer->s->name) <= BALANCER_PREFIX_LENGTH ||
                 strcasecmp(&balancer->s->name[BALANCER_PREFIX_LENGTH], node->mess.balancer) != 0) {
                 continue;
             }


jfclere (commented):
Should we have
if (strlen(balancer->s->name) <= BALANCER_PREFIX_LENGTH) {
  continue;
}
if (strcasecmp(&balancer->s->name[BALANCER_PREFIX_LENGTH], node->mess.balancer) != 0) {
  continue;
}
The test for strlen() is to check that the strcasecmp is not comparing somewhere outside the balancer->s->name... Is that even possible?

jajik:
> Should we have if (strlen(balancer->s->name) <= BALANCER_PREFIX_LENGTH) { continue; } if (strcasecmp(&balancer->s->name[BALANCER_PREFIX_LENGTH], node->mess.balancer) != 0) { continue; } The test for strlen() is to check that the strcasecmp is not comparing somewhere outside the balancer->s->name... Is that even possible?

The proposed change is equivalent to the two ifs you wrote if I read that correctly.

For the second part I don't think it's possible because `BALANCER_PREFIX_LENGTH` is smaller than the array size. However, I thought that the `strlen` should prevent reading invalid balancer names (e.g. if we have a valid balancer name for which we change the first char to `\0`, we would catch it with the `strlen` but wouldn't without it).

jfclere (approved):

Issue #265, opened 2024-08-13
Handle `nodeinfo.offset` better


jajik:
1) The `offset` is not necessary to keep in each instance, it is a compile time constant.
    - Done
2) Fix offset computation (`unsigned long` vs `int`). 
    - Done
3) Additionally, is the `offset` needed at all? Are there configurations where `nodeinfo.stat` won't be aligned?

jajik:
After some investigation, I think the offset is needed because of a little more exotic architectures (see [MODCLUSTER-831(https://issues.redhat.com/browse/MODCLUSTER-381)). Basically, we want to preserve 64-bit alignment even on 32-bit systems, so this is indeed needed.

What we might consider is changing `SIZEOFSCORE` which seems unnecessarily big (currently, it's 1700, the `proxy_worker_shared` structure has less than 1500 bytes on my system – imho we might do something like `#define SIZEOFSCORE ALIGNED_PROXY_WORKER_SHARED_SIZE + 8` or so).

fyi @jfclere

rhusar:
@jajik Perhaps add a source code comment to describe this for future coders?

jajik:
That's not a bad idea. Let's wait for JFC and do it with the possible change of `SIZEOFSCORE`.

PR #270, opened 2024-08-16
Correct the offset computation

jajik:
Partly addresses #265, but we still should probe whether the offset is necessary at all.

### DIFF
--- a/native/mod_manager/node.c
+++ b/native/mod_manager/node.c
@@ -88,7 +88,7 @@ static apr_status_t update(void *mem, void *data, apr_pool_t *pool)
         in->mess.id = ou->mess.id;
         memcpy(ou, in, sizeof(nodemess_t));
         ou->updatetime = apr_time_now();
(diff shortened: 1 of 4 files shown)


native/mod_manager/node.c
```diff
@@ -78,18 +78,8 @@ static apr_status_t update(void *mem, void *data, apr_pool_t *pool)
...
-        /*
-         * The node information is made of several pieces:
-         * Information from the cluster (nodemess_t).
-         * updatetime (time of last received message).
-         * offset (of the area shared with the proxy logic).
-         * stat (shared area with the proxy logic we shouldn't modify it here).
-         */
         in->mess.id = ou->mess.id;
         memcpy(ou, in, sizeof(nodemess_t));
-        ou->updatetime = apr_time_now();
```
jajik, line 90: @jajik  This is an error.

jajik: Fixed.

jfclere (approved):

jajik:
Thanks for the review.

PR #271, opened 2024-08-20
MODCLUSTER-822: Decline Contexts and Aliases if lengths are exceeded

jajik:
ref [MODCLUSTER-822](https://issues.redhat.com/browse/MODCLUSTER-822)

This approach abstracts processing of Alias and Context for both APP and CONFIG commands. As a result, Alias for CONFIG is now also case insensitive, but that seems more like the correct behavior...


jfclere (approved):
Looks OK for me.

jfclere:
Any plans to add tests for the new messages? ;-)

jajik:
> Any plans to add tests for the new messages? ;-)

I hope that I'll manage to add the httpd/perl framework soon and write tests in that. I added this PR to the Missing tests issue.

jajik:
Merging. Thanks for the review!

PR #273, opened 2024-08-29
Do not accept CMD parameters without a value in manager

jajik:
e.g. CONFIG msg with `Host=`. This change causes such message to be declined with the same error as is used in 1.3.x.


jfclere (approved):
thanks good catch.

PR #274, opened 2024-08-29
test: Tweak test functions names, drop unnecessary ones

jajik:
I got annoyed with the inconsistency & also `tomcat_stop` is useless so I removed it.


rhusar:
Thanks @jajik

PR #275, opened 2024-08-30
Fix mod_manager.c DUMP command output

jajik:
Fix the output. Currently, one would get

```
node: [1:1],Balancer: mycluster,JVMRoute: spare,LBGroup: [],Host: localhost,Port: 6666,                           Type: ws,flushpackets: 0,flushwait: 10,ping: 10,smax: -1,ttl: 60,timeout: 0
```

(see the gap before `Type`)

instead of

```
node: [1:1],Balancer: mycluster,JVMRoute: spare,LBGroup: [],Host: localhost,Port: 6666,Type: ws,flushpackets: 0,flushwait: 10,ping: 10,smax: -1,ttl: 60,timeout: 0
```

after the fix.

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -1646,17 +1646,18 @@ static char *process_dump(request_rec *r, int *errtype)
 
         switch (type) {
         case TEXT_XML:
-            ap_rprintf(r, "<Balancer id=\"%d\" name=\"%.*s\">\
-                           <StickySession>\
-                               <Enabled>%d</Enabled>\
-                               <Cookie>%.*s</Cookie>\
-                               <Path>%.*s</Path>\
-                               <Remove>%d</Remove>\
-                               <Force>%d</Force>\
-                           </StickySession>\
-                           <Timeout>%d</Timeout>\
-                           <MaxAttempts>%d</MaxAttempts>\
-                           </Balancer>",
+            ap_rprintf(r,
+                       "<Balancer id=\"%d\" name=\"%.*s\">"
+                       "<StickySession>"
+                       "<Enabled>%d</Enabled>"
+                       "<Cookie>%.*s</Cookie>"
+                       "<Path>%.*s</Path>"
+                       "<Remove>%d</Remove>"
+                       "<Force>%d</Force>"
+                       "</StickySession>"
+                       "<Timeout>%d</Timeout>"
+                       "<MaxAttempts>%d</MaxAttempts>"
+                       "</Balancer>",
                        id[i], (int)sizeof(ou->balancer), ou->balancer, ou->StickySession,
                        (int)sizeof(ou->StickySessionCookie), ou->StickySessionCookie,
                        (int)sizeof(ou->StickySessionPath), ou->StickySessionPath, ou->StickySessionRemove,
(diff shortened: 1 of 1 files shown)


jajik:
I've just added the same fix for XML, that looks like this otherwise

```
<?xml version="1.0" standalone="yes" ?>
<Dump><Balancers><Balancer id="0" name="mycluster">                           <StickySession>                               <Enabled>1</Enabled>                               <Cookie>JSESSIONID</Cookie>                               <Path>jsessionid</Path>                               <Remove>0</Remove>                               <Force>1</Force>                           </StickySession>                           <Timeout>0</Timeout>                           <MaxAttempts>1</MaxAttempts>                           </Balancer></Balancers><Nodes><Node id="0">                               <Balancer>mycluster</Balancer>                               <JVMRoute>fake</JVMRoute>                               <LBGroup></LBGroup>                               <Host>localhost</Host>                               <Port>8888</Port>                               <Type>ws</Type>                               <FlushPackets>0</FlushPackets>                               <FlushWait>10</FlushWait>                               <Ping>10</Ping>                               <Smax>-1</Smax>                               <Ttl>60</Ttl>                               <Timeout>0</Timeout>                           </Node><Node id="1">                               <Balancer>mycluster</Balancer>                               <JVMRoute>spare</JVMRoute>                               <LBGroup></LBGroup>                               <Host>localhost</Host>                               <Port>6666</Port>                               <Type>ws</Type>                               <FlushPackets>0</FlushPackets>                               <FlushWait>10</FlushWait>                               <Ping>10</Ping>                               <Smax>-1</Smax>                               <Ttl>60</Ttl>                               <Timeout>0</Timeout>                           </Node></Nodes><Hosts></Hosts><Contexts></Contexts></Dump>
```

Unfortunately that applies to both versions so I'll probably create a JIRA for it.

rhusar (commented):
So this is a regression introduced in your previous commit - https://github.com/modcluster/mod_proxy_cluster/commit/edd79e4bdb8aba4243dededfdaa14e93972632f3#diff-6bf23800d3aab4e39caab971522dc2c1942641627e3f7d2a2f28ae97fa74c038L1527

Is this a bug in the tooling you used for the change (I assume clang)?

jajik:
> So this is a regression introduced in your previous commit - [edd79e4#diff-6bf23800d3aab4e39caab971522dc2c1942641627e3f7d2a2f28ae97fa74c038L1527](https://github.com/modcluster/mod_proxy_cluster/commit/edd79e4bdb8aba4243dededfdaa14e93972632f3#diff-6bf23800d3aab4e39caab971522dc2c1942641627e3f7d2a2f28ae97fa74c038L1527)

The first commit, yes, it seems so.

> Is this a bug in the tooling you used for the change (I assume clang)?

I doubt it. I would guess the most probable explanation is that I got tricked by the other strings. 

-----

What do you think about the XML output?

jajik:
For the record, we agreed to go with the XML without any additional formatting as proposed in this PR.

rhusar (approved):
LGTM

rhusar:
Thanks @jajik 

Doesn't effect downstream, no Jira needed.

rhusar:
> For the record, we agreed to go with the XML without any additional formatting as proposed in this PR.

That's right, no need for redundant whitespace.

jajik:
> Thanks @jajik
> 
> Doesn't effect downstream, no Jira needed.

That's not true regarding the XML part. I'll create one eventually.

rhusar:
> > Thanks @jajik
> > Doesn't effect downstream, no Jira needed.
> 
> That's not true regarding the XML part. I'll create one eventually.

I am not sure we need one as that logic is never used.

PR #280, opened 2024-09-11
Remove unsupported methods from the codebase

jajik:
These methods are undocummented, without an implementation, and were never used


rhusar (approved):
The add remove and query I think are from the POC implementation for jgroups discovery. But that has been abandoned and code purged. So I think this is a remnant of that. But I defer to @jfclere

jfclere:
@rhusar yes those were for some testing of some jgroups stuff. They should be removed now.

jfclere (approved):

rhusar:
Great, thanks @jfclere and @jajik !

PR #283, opened 2024-10-02
Add an initial support of Apache::Test framework

### LINKED ISSUE #130: Testsuite: integrate a proper testing framework rather than using set of custom scripts

jajik:
This PR adds a very limited set of tests that use httpd's perl test framework. These tests are run in CI as well.

close #130


jajik:
Worth of noting, the structure of CI will become slightly more complicated because `perl-tests` depend on `make-with--httpd-trunk`, but that imho does not matter.

The reason of this solution is that httpd in Ubuntu has a weird configuration that did not go well with the testing framework. I thought about the possibility of using fedora image but then we would have to complicate things with non-root user and switching users throughout the workflow and stuff like that. Using already built httpd with the modules seems appropriate (and we don't have to build them again as we would have for the aforementioned solutions).

jfclere (approved):

Issue #285, opened 2024-10-04
Improve handling of received parameters in mod_manager.c


jajik:
There are commands whose parameters may change settings of the proxy but their processing repeats code and checks only non-default options.  E.g.

```c
if (strcasecmp(key, "Reversed") == 0) {
    if (strcasecmp(val, "yes") == 0) {
        nodeinfo->mess.reversed = 1;
    }
}
```

We should update this approach with some appropriate abstraction covering all options.

jajik:
To be clear, for example requests with following options `Reversed=yes&Reversed=no` are not handled properly.

PR #287, opened 2024-10-09
Fix and improve handling boolean parameters

jajik:
This fixes edge cases where the current solution did not work properly, but it also introduces an equivalence between `yes`/`no` and `on`/`off`. The latter can be changed so that the behavior is the same, but I think it's reasonable...


jajik:
close #285

jfclere (approved):

PR #290, opened 2024-10-17
Improve handling of overlapping aliases in APP commands

jajik:
Previously, an app would get merged into an existing vhost only in case the first alias would be already in. So that VHost with alias `test` would be updated for command with `Alias=test,demo` but a new VHost would be created for `Alias=demo,test`.

This PR changes the behavior. Now we go through all provided aliases sequentially, and if one matches, the corresponding VHost gets updated. Otherwise a new one is created.

Resolves [MODCLUSTER-824](https://issues.redhat.com/browse/MODCLUSTER-824)

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -2084,7 +2084,7 @@ static char *process_appl_cmd(request_rec *r, char **ptr, int status, int *errty
 
     int i = 0;
     hostinfo_t hostinfo;
-    hostinfo_t *host;
+    hostinfo_t *host = NULL;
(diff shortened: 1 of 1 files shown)


jfclere (approved):

jfclere:
I don't see how to trigger the problem adding a war in a running tomcat... But the fix makes sense.

rhusar:
> I don't see how to trigger the problem adding a war in a running tomcat... But the fix makes sense.

IIUC it's fairly reare, but not impossible situation, no? Reconfigure the server to add a new alias, kill the server, start it again.

Issue #292, opened 2024-10-25
Make perl tests required in CI


jajik:
[ ] - "fake" CI job must be introduces
[ ] - settings of the repository must change

rhusar:
perl tests are now required for main branch ✅

jajik:
That was fast! Thanks, the PR is up.

rhusar:
Done, closing. T hannks

PR #293, opened 2024-10-25
ci: Add fake perl-tests job to ignored workaround

jajik:
ref #292


rhusar:
Exactly the reason why I didn't really like this solution, oh well.. :)

rhusar:
Thanks @jajik

PR #294, opened 2024-10-25
Add a test for MODCLUSTER-824

jajik:
and tweak the rest to have a clean table after each test...

Note, that these sleeps are needed because it takes some time for a node to disapear (it must go through Node REMOVED phase). I'll try to come up with something more elegant in the future.


jajik:
ref the original PR #290

jfclere (approved):

PR #296, opened 2024-10-31
Fix an error message to contain the correct directive name

jajik:
There is a typo/error in the error message: *for* -> *before*.


jfclere (approved):

jajik:
Thanks for the review

rhusar:
Interestingly enough, the https://docs.modcluster.io/legacy/ mention the old variant but https://docs.modcluster.io/#waitbeforeremove is correct.

jfclere:
@rhusar do you want to support both version? or assume we are good?

jajik:
I've just quickly scanned the repo and it seems the directive was always `WaitBeforeRemove` and that the docs were simply wrong. At least that's how it looks like.

PR #299, opened 2024-11-05
ci: Remove branches '*' when it's the default behaviour

jajik:
Well, that's not completely true. The default behaviour is `'**'`, the single `*` does not match branches with `/` in their name. Fortunately, we are all well behaved and don't use such names. :slightly_smiling_face: 

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

### DIFF
--- a/.github/workflows/ci-ignored-workaround.yml
+++ b/.github/workflows/ci-ignored-workaround.yml
@@ -1,15 +1,11 @@
 name: Skip CI on ignored paths
 on:
   push:
-    branches:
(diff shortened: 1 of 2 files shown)


rhusar:
I am kinda surprised this is not required.. but if that works and you want it, why not.

rhusar:
I guess it's because we added 'paths' since then and you can't have an empty tree in yml.

PR #302, opened 2024-11-07
Add `UseNocanon` to mod_lbmethod_cluster, fix testsuite to use the right module

### LINKED ISSUE #297: Fix the testsuite for mod_proxy_balancer
The `MODCLUSTER-*` tests with mod_proxy_balancer does not work because they use their custom configs with mod_proxy_cluster.

### LINKED ISSUE #298: Add support for `UseNocanon` to mod_lbmethod_cluster
Seems reasonable to allow configuring that.

ref #297

jajik:
close #297 
close #298 

This resolves two issues, but given that they are connected it seemed reasonable to do it at once. Test that ought to be run with balancer were unfortunately run with cluster. The testsuite needed a few changes and new conf files for the second module.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,7 +244,7 @@ jobs:
       - name: Setup dependencies
         run: |
           sudo apt-get update -y
-          sudo apt-get install -y docker maven git curl iproute2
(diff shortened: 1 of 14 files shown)


jfclere (approved):

jajik:
Thanks for the review

PR #305, opened 2024-11-20
Fix for https://issues.redhat.com/browse/MODCLUSTER-774

jfclere:
That prevents the message:
 AH00898: URI cannot be parsed: ://:0/ returned by /

### DIFF
--- a/native/mod_proxy_cluster/mod_proxy_cluster.c
+++ b/native/mod_proxy_cluster/mod_proxy_cluster.c
@@ -201,6 +201,11 @@ static void add_hcheck(server_rec *s, const proxy_server_conf *conf, proxy_worke
 
 static int (*ap_proxy_retry_worker_fn)(const char *proxy_function, proxy_worker *worker, server_rec *s) = NULL;
 
+static int is_worker_empty(proxy_worker *worker)
+{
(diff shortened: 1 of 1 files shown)


jajik (changes requested):
Given that the same check is already used [here](https://github.com/modcluster/mod_proxy_cluster/blob/44dd11e89bfadaa7dad62c7536daff44cd91c649/native/mod_proxy_cluster/mod_proxy_cluster.c#L235), I would suggest adding a helper function

```c
int is_worker_empty(const proxy_worker *worker) {
    return worker && (worker->s->port == 0 || worker->s->scheme[0] == '\0' || worker->s->hostname[0] == '\0');
}
```

and used that instead (in both cases)-

jajik (changes requested):
It will need

`clang-format -i mod_proxy_cluster/mod_proxy_cluster.c`

and I'm not sure about the `volatile` keyword we have there (isn't it always under lock?) I'll think about it..

jajik:
I looked at it again and I think we can drop the `volatile` from `volatile proxy_worker *worker = *workers;` as well as from the function signature.

jajik (approved):
LGTM, and if you squash the middle commits, it will be perfect

jajik:
Rebasing and I'll merge it afterwards

PR #307, opened 2024-12-04
test: Use default configuration for tomcat containers

jajik:
There is no reason to use non-default values.


test/tomcat/server.xml
```diff
@@ -24,10 +24,6 @@
             advertise="true" stickySession="true" stickySessionForce="false" stickySessionRemove="true" /> -->
   <Listener className="org.jboss.modcluster.container.catalina.standalone.ModClusterListener"
             connectorPort="8080"
-            advertise="false"
-            stickySession="true"
-            stickySessionForce="false"
-            stickySessionRemove="true"
```
rhusar, line 30: FYI - the stickySessionRemove is false by default. But clearly that's not tested anywhere here.

rhusar (approved):

rhusar:
Thanks @jajik

PR #311, opened 2024-12-19
test: Update pom.xml to use the latest mod_cluster snapshot

jajik:


rhusar (approved):
LGTM

n.b. while this should really be a fixed version, it's pretty great that we get continuous testing of main. If this ever becomes a problem we can change that.

PR #314, opened 2025-01-14
Improve Makefiles for modules, use environment variables through GH Actions

jajik:


.github/workflows/ci.yml
```diff
@@ -118,7 +120,7 @@ jobs:
           for module in advertise/ mod_proxy_cluster/ balancers/ mod_manager/; do \
             cd $module; \
             sh buildconf; \
-            ./configure CFLAGS="-Wall -Werror" --with-apxs=/usr/local/apache2/bin/apxs; \
+            ./configure --with-apxs=/usr/local/apache2/bin/apxs; \
```
rhusar, line 123: Sorry but can you explain how this is an improvement? As someone who has already done loads of CI environment migrations, I don't favor employing more vendor-locking for no value, but perhaps I am missing the benefit here?

jajik: Personally, I find it more clean.

But there are two things: 1) how to build it; 2) which flags to use. For the first one, custom flags are not relevant and the copy-pasting gets clearer. For the second one, the custom flags are relevant only in debugging the pipeline locally (in case of a failure), but in that case, one should look at the whole job, so I don't think it's an issue.

Additionally, it's possible to supply flags to both, `configure` and `make` with different implications. I would prefer people to use the one they need instead of simply copy-pasting.

With that said, it's not a hill I would die on.

rhusar: OK, let's do it then.

PR #315, opened 2025-01-14
Add coverage reporting support for the testsuite

### LINKED ISSUE #282: Explore how to add code coverage tests/reporting to the repository

jajik:
Opening as a draft because I think the generated report is incomplete (well, I guess the build is not executed properly and because of that not everything is marked as executed). I just want to have it out there.

Both lcov & gcov reports are generated, but the source data are the same.


rhusar:
I think this is starting to look good! I will be curious to review the reports.

jajik:
Except it's not reported properly. I'll talk to the compiler a little bit more, maybe he'll change his mind... :slightly_smiling_face:

jajik:
It's my pleasure to make this ready for review. The coverage is working (at least it seems so) now.

I used multiple AIs for debugging my original solution, because the coverage reported was low and clearly incorrect (functions that must have been executed in order for tests to pass were marked as not executed). The best results I got were from Claude, which overengineered a "fix" that worked. It was needlessly complicated, so I didn't use it; however, it pointed out the issue I had in my solution – missing permissions for the workers, which are fixed using `chmod` with `umask`. It was an interesting exercise though. :slightly_smiling_face:

Main changes are:

* code coverage reporting for tests
* by default ON in CI (-> coverage artifacts with html reports)
* tests compile modules using CMake now

PR #316, opened 2025-01-15
Add shields.io badge and link to RPM package in Fedora to the README …

rhusar:
…file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-mod_cluster
+mod_cluster [![Fedora package](https://img.shields.io/fedora/v/mod_proxy_cluster)](https://packages.fedoraproject.org/pkgs/mod_proxy_cluster/mod_proxy_cluster/)
 ===========
 
 Project mod_cluster is a httpd-based load-balancer. It uses a communication channel to forward


rhusar:
This is what it looks like btw - https://github.com/rhusar/mod_proxy_cluster/tree/shieldsio-rpm?tab=readme-ov-file#mod_cluster-

rhusar:
@jajik Can you confirm we want to link to mod_proxy_cluster RPM package and not the mod_cluster one?

jajik:
> @jajik Can you confirm we want to link to mod_proxy_cluster RPM package and not the mod_cluster one?

Currently, it does not matter that much, both packages are on 1.3.21.Final. After 2.0.0.Final gets released, I'll update mod_proxy_cluster only (leaving mod_cluster package on 1.3.x version) so yes, linking mod_proxy_cluster is probably better if we're fine with referencing code that is outside of this repository in the meantime.

jajik (approved):

rhusar:
Also, the scope of the original RPM (mod_cluster) seems to have been above the Apache httpd module, asl intended to bring in tomcat Java integration libraries...

rhusar:
Thx @jajik

PR #317, opened 2025-01-15
Fix CMakeLists by specifying LANGUAGES explicitly, update minimal version

jajik:
If `LANGUAGES` are not specified, C and C++ are the default. Which forces users to have a C++ compiler or use additional flags. We don't need that. 

The minimum CMake version is almost 9 years old so it shouldn't be an issue and we'll get rid of some warnings.


jfclere (commented):
The version I have on windoze test box is too old to test:
+++
C:\Users\Administrator>cmake --version
cmake version 3.22.22022201-MSVC_2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
+++

jajik:
> The version I have on windoze test box is too old to test: +++ C:\Users\Administrator>cmake --version cmake version 3.22.22022201-MSVC_2
> 
> CMake suite maintained and supported by Kitware (kitware.com/cmake). +++

Are you sure? It seems you have 3.**22** which should be fine.

jfclere:
> > The version I have on windoze test box is too old to test: +++ C:\Users\Administrator>cmake --version cmake version 3.22.22022201-MSVC_2
> > CMake suite maintained and supported by Kitware (kitware.com/cmake). +++
> 
> Are you sure? It seems you have 3.**22** which should be fine.

I had something wrong in my VM I have built using Visual Studio 17 2022

jfclere (approved):

jajik:
Thanks for the review!

PR #322, opened 2025-02-19
Use Containerfile naming, tweak & polish these, clean up unused files 

jajik:
Dropping Dockerfile naming so that we're consistent with other repos (e.g. https://github.com/modcluster/ci.modcluster.io/).


rhusar (approved):

rhusar:
Thanks @jajik

PR #323, opened 2025-02-25
test: Update *_create functions to use Containerfile explicitly

jajik:
Apparently, docker has an issue with Containerfiles unless you specify it with `-f` option. We've started using `podman` internally a few PRs ago, so it does not manifest here in our CI, but on systems with `docker` it doesn't work.


jajik:
ref https://github.com/modcluster/mod_proxy_cluster/pull/322

rhusar:
Thanks @jajik

PR #324, opened 2025-03-18
ci: Run maintainer-mode job on the latest and the oldest supported httpd

### LINKED ISSUE #321: Consider running CI with the oldest supported httpd as well
Since we're using always the latest we might run into a situation when we'll break backwards compatibility without realizing.

It might be the case in situations like [this](https://github.com/modcluster/mod_cluster/pull/873) (it is not in this case).

jajik:
close #321 

The solution uses a new container image (it's [the same one](https://github.com/modcluster/ci.modcluster.io/tree/main/httpd-dev) we already use but with a different version of httpd – 2.4.53). This image will be managed manually as the rebuild is necessary only in case the oldest supported version changes.

edit: Adding link https://quay.io/repository/mod_cluster/ci-httpd-dev?tab=tags

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,8 +104,12 @@ jobs:
 
   make-httpd-maintainer-mode:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image-tag: [ "latest", "oldest-supported" ]
     container:
-      image: quay.io/mod_cluster/ci-httpd-dev
+      image: quay.io/mod_cluster/ci-httpd-dev:${{ matrix.image-tag }}
     defaults:
       run:
         shell: bash
@@ -114,6 +118,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Print httpd version
+        run: echo $HTTPD_VERSION
       - name: Build
         run: |
           cd native


.github/workflows/ci.yml
```diff
@@ -104,8 +104,12 @@ jobs:
 
   make-httpd-maintainer-mode:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image-tag: [ "latest", "oldest-supported" ]
     container:
-      image: quay.io/mod_cluster/ci-httpd-dev
+      image: quay.io/mod_cluster/ci-httpd-dev:${{ matrix.image-tag}}
```
rhusar, line 112: Format only, missing space before `}}`

.github/workflows/ci.yml
```diff
@@ -104,8 +104,12 @@ jobs:
 
   make-httpd-maintainer-mode:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image-tag: [ "latest", "oldest-supported" ]
```
rhusar, line 110: BTW I changed my opinion on using arrays and I prefer itemizations now as these can be commented out per member or changed without changing the whole line containing all options. But decide as you prefer, e.g. for things that do never change, it doesn't matter.

rhusar (changes requested):

.github/workflows/ci.yml
jajik: We have it in a few places. I don't really care, but if you do, I would change everywhere.

rhusar (approved):
Excellent, I think we really needed this!

rhusar:
Thanks @jajik

jajik:
Thanks for the review.

PR #327, opened 2025-03-31
test: Add timeout for curls, minor syntax tweaks

### LINKED ISSUE #326: Add timeout to all curls used in the testsuite
When the timeout is missing and there is an error (typicaly a deadlock) while processing the request, tests get stuck. We should use `curl -m <value>` in all cases.

jajik:
With a timeout set, tests shouldn't get stuck waiting for a response.

Also I've changed `\`...\`` to `$(...)` syntax on a few places to be consistent with the rest.

close #326


rhusar (approved):
LGTM

rhusar:
Thanks @jajik

PR #328, opened 2025-04-03
Add missing balancer sync to lbmethod_cluster, minor tweaks

### LINKED ISSUE #325: Fix (still) unstable MODCLUSTER-640 test
The test is still unstable and occasionally fails. (See for example [this](https://github.com/jajik/mod_proxy_cluster/actions/runs/13950351891)).

In very rare cases there's some segmentation fault issue which is hard to trigger. In most cases nodes are changing their statuses (OK/NOTOK) for some reason and the number of OK nodes does not meet the expectations. The changing should not happen though.

jajik:
The missing sync was the problem I was looking for. However, there are a few additional tweaks.

close #325


jajik:
Thanks a lot for the review!

Issue #330, opened 2025-06-06
Take a look at error messages in case of invalid MCMP commands


jajik:
Currently there are some cases that are not handled well.

For example

* `CONFIG` with no parameters results in an error `SYNTAX: JVMRoute can't be empty`
* `CONFIG` with a single parameter `jvmroute=` results in an error `SYNTAX: Can't parse MCMP message. It might have contained illegal symbols or unknown elements.`

That doesn't look very intuitive.

(Note: version 1.3.x handled both cases with the later message.)

Issue #329, opened 2025-05-22
Improper handling of multiple contexts


jajik:
Contexts are not properly handled in case they are provided in separate parameters. That means that `context=/one,/two` is not handled the same way as `context=/one context=/two` even though this works for aliases.

PR #331, opened 2025-06-08
Fix handling of duplicite MCMP command parameters 

jajik:
This addresses #329 and adds additional tests for it. However, there is an error message change for which I opened #330. (tl;dr This unifies the behaviour of 2.x with 1.3.x but both are inaccurate.)


jajik:
@rhusar Could you please eventually change the required checks so that we can merge it?

rhusar:
already changed some but it looks like i missed these. Will fix.

rhusar:
Thanks @jajik @jfclere

PR #332, opened 2025-06-10
test: Tweak test containers, reduce size

jajik:
Also "fixed" `ADD`/`COPY` usage.

### DIFF
--- a/test/httpd/Containerfile
+++ b/test/httpd/Containerfile
@@ -1,22 +1,22 @@
-FROM fedora:41
+FROM fedora:41 AS builder
 
 ARG HTTPD_SOURCES="https://dlcdn.apache.org/httpd/httpd-2.4.63.tar.gz"
 
-RUN yum install gcc wget apr-devel apr-util-devel openssl-devel pcre-devel redhat-rpm-config wcstools git autoconf -y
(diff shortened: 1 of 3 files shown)


test/httpd/Containerfile
```diff
@@ -1,22 +1,24 @@
-FROM fedora:41
+FROM fedora:41 AS builder
 
 ARG HTTPD_SOURCES="https://dlcdn.apache.org/httpd/httpd-2.4.63.tar.gz"
 
-RUN yum install gcc wget apr-devel apr-util-devel openssl-devel pcre-devel redhat-rpm-config wcstools git autoconf -y
+RUN dnf install pcre apr-util wcstools -y
+
+RUN dnf install gcc apr-devel apr-util-devel openssl-devel pcre-devel redhat-rpm-config autoconf -y
```
rhusar, line 7: Why two separate RUNs?

jajik: It can be used by the subsequent stage – shared layer (in the end it doesn't matter that much).

rhusar: I guess I don't see why we would start dnf twice, load the cache twice, resolve dependency tree twice, etc.

jajik: Done.

rhusar (approved):

rhusar:
Thanks @jajik !

PR #334, opened 2025-06-10
Removed unused messages after jgroups discovery removal #333

rhusar:


jajik:
Maybe *Remove* would be better than  *Removed* but that's not important.

rhusar:
> Maybe _Remove_ would be better than _Removed_ but that's not important.

Yeah, it doesn't change the intention, but as you know I prefer present form (this was a typo indeed) so the commit messages can be vied as 'a command to the code base'. Fixed.

jajik:
Merging. Thank you!

PR #337, opened 2025-06-27
Use only port 8000 in tests

jajik:
For some (historic?) reasons we've been using two ports in tests, so let's use just one instead.


jajik:
We're currently using two (and we've been since I remember) and I usually accessed through `6666`. But frankly I don't care, we can change it to any other one, I'd just like to end using two ports. Any suggestions or wishes?

jajik:
@rhusar I changed the port to the other we are using (8000).

rhusar (changes requested):
You are not going to like this, but the 8000 is used by tomcat as the default JDPA port. It can be changed to any port, but this is default, which makes little sense to have conflicting ports in a default setup.

rhusar:
> BTW Here's the Jira for this https://issues.redhat.com/browse/MODCLUSTER-858

Docs part for this: https://github.com/modcluster/mod_proxy_cluster/pull/337#issuecomment-3078357410

PR #342, opened 2025-08-04
Fix ENV variable declaration in the test container, update fedora version to 42, tweak CI

jajik:
The issue with the env variable (in the Containerfile) does not manifest in the testsuite because we're using a default in `httpd_start`, but still.


jajik:
Added a fix for our CI workflows, let's use only `apt-get` and not mix it with `apt`. Additionally, I removed duplicates in packages list and an unnecessary flag.

PR #352, opened 2025-09-11
test: Use mod_cluster zips, improve handling of dependencies

jajik:


rhusar (approved):
OK, this is fine, moves the tests to use the zip distribution. 

Nevertheless, those should probably be downloaded from maven directly, but ideally, we would want to publish snapshots to maven for this. Anyway, we have a pin in that.

rhusar:
Thanks @jajik

PR #357, opened 2025-11-11
Tests: Use custom docker/podman network, rework provisioning of containers, and more tweaks

### LINKED ISSUE #348: Use custom docker network instead of host in the testsuite
* by default, only proxy port should be exposed
* additionally, we might use custom hostnames for free instead of `localhost`

jajik:
This is a bigger PR but because of the nature of changes, there is not much what to do about it (I've already dropped some changes for future PRs).

The main thing is that host network is no longer used in containerized tests. This allows us to use DNS, isolate tests to a different network (minimizing conflicts with other running services).

In points:

* uses docker/podman custom network instead of host network
* leverages DNS instead of IPs/localhost
* separates UseAlias test from maintests (and UseAlias is not enabled by default in the rest of the tests)
* reworks and improves containers provisioning (gets rid of some old limitations, makes configurable things that make sense, drops support for things we were not using)
* minor tweaks and fixes (POSIX operators, drop unnecessary/failing commands, tweak log levels, fix typos, etc.)

close #348

### DIFF
--- a/test/MODCLUSTER-734/testit.sh
+++ b/test/MODCLUSTER-734/testit.sh
@@ -9,10 +9,7 @@ httpd_remove
 # build httpd + mod_proxy_cluster
 rm -f nohup.out
 
-MPC_CONF=${MPC_CONF:-MODCLUSTER-734/mod_proxy_cluster.conf}
-MPC_NAME=MODCLUSTER-734
-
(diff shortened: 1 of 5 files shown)


jajik:
Thanks for the review!

PR #365, opened 2026-01-08
test: Make jdbsuspend helper functions available within testsuite

jajik:
With this, suspend becomes widely available in the whole testsuite (and even outside of it, so that I can subsequently use it in the perf suite without touching any sources).


jfclere (approved):
thanks

rhusar:
Thanks @jajik and @jfclere !

PR #367, opened 2026-01-30
ci: Run tests on ubuntu arm64 runner

jajik:
Ubuntu ARM runner is available, so I don't see why not to enable this as well.


jajik:
@rhusar This might require changing the required/expected checks (as it splits `tests` into `tests (ubuntu-latest)` and `tests (ubuntu-24.04-arm)`).

PR #377, opened 2026-03-20
Fix id for PING and STATUS responses

jajik:
`restart_time` is of type `apr_int64_t`, so `(int)` is not correct (it leads to ugly numbers)


rhusar:
@jfclere Please review.

rhusar (changes requested):
Using %ld is not fully portable. The fix should use APR_INT64_T_FMT instead.

See native/advertise/mod_advertise.c for example.

jajik:
> Using %ld is not fully portable. The fix should use APR_INT64_T_FMT instead.
> 
> See native/advertise/mod_advertise.c for example.

Good point. Fixed everywhere in the latest commit.

jfclere (approved):
Good thanks

jajik:
@rhusar the change you requested was applied.

PR #378, opened 2026-03-24
Prevent wrong values for load

jajik:
This should be also backported: [MODCLUSTER-869](https://redhat.atlassian.net/browse/MODCLUSTER-869). (TBH the whole function/logic needs refactor, but I'll do that later.)

### DIFF
--- a/native/mod_manager/mod_manager.c
+++ b/native/mod_manager/mod_manager.c
@@ -2424,6 +2424,10 @@ static char *process_status(request_rec *r, const char *const *ptr, int *errtype
         i++;
     }
 
+    if (Load < -1 || Load > 100) {
+        ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, r->server, "Load value %d is outside the supported range", Load);
+    }
+
     /* Read the node */
     loc_lock_nodes();
     node = read_node(nodestatsmem, &nodeinfo);
--- a/native/mod_proxy_cluster/mod_proxy_cluster.c
+++ b/native/mod_proxy_cluster/mod_proxy_cluster.c
@@ -1833,7 +1833,7 @@ static proxy_worker *internal_find_best_byrequests(const proxy_balancer *balance
  * Check that we could connect to the node and create corresponding balancers and workers.
  * id   : worker id
  * load : load factor from the cluster manager.
- * load > 0  : a load factor.
+ * load between 1 and 100: a load factor.
  * load = 0  : standby worker.
  * load = -1 : errored worker.
  * load = -2 : just do a cping/cpong.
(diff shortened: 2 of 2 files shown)


rhusar:
@jfclere I don't think the lack of the check was intentional, but please review.

rhusar (changes requested):
Consider adding a debug-level log for out-of-range values to aid troubleshooting.

jajik:
> Consider adding a debug-level log for out-of-range values to aid troubleshooting.

Fixed.

jajik:
(just a rebase)

native/mod_manager/mod_manager.c
```diff
@@ -2424,6 +2424,10 @@ static char *process_status(request_rec *r, const char *const *ptr, int *errtype
         i++;
     }
 
+    if (Load < -1 || Load > 100) {
```
rhusar, line 2427: Doesn't this need to handle -2 used here?

https://github.com/modcluster/mod_proxy_cluster/pull/378/changes#diff-7fb9cb6f7b5039b15e6d91f8624efdffd1d66bf388ab373834e6acb7a17dac99R1889

jfclere (approved):
Looks good to me.

native/mod_manager/mod_manager.c
jajik: No, the `-2` is an internal hack for `process_ping` https://github.com/jajik/mod_proxy_cluster/blob/2a598b776795cd5dd5ae5c50fd6ccded0449d114/native/mod_manager/mod_manager.c#L2549 so it won't get triggered here (but if anybody sends a status with `-2`, we want the log).

jajik:
@rhusar Requested changes are in. Do you have anything else, or can we merge?

rhusar (changes requested):
@jajik -2 can still reach process_status from an external caller. The Load value is parsed via bare atoi() with no validation (mod_manager.c:2418), so anyone sending STATUS ... Load=-2 over the wire will have it accepted and passed to proxy_node_isup() at line 2443 — the same code path as process_ping.

The new check if (Load < -1 || Load > 100) only logs a debug message but does not reject or clamp the value. So after the log, -2 is still passed through to proxy_node_isup(), which will perform a cping/cpong and return early without updating worker status — effectively letting an external STATUS command act as a PING.

The comment doesn't covert this either

```
/*
 * Process the STATUS command
 * Load -1 : Broken
 * Load 0  : Standby.
 * Load 1-100 : Load factor.
 */
static char *process_status(request_rec *r, const char *const *ptr, int *errtype)
{
```

jajik:
> @jajik -2 can still reach process_status from an external caller. The Load value is parsed via bare atoi() with no validation (mod_manager.c:2418), so anyone sending STATUS ... Load=-2 over the wire will have it accepted and passed to proxy_node_isup() at line 2443 — the same code path as process_ping.
> 
> The new check if (Load < -1 || Load > 100) only logs a debug message but does not reject or clamp the value. So after the log, -2 is still passed through to proxy_node_isup(), which will perform a cping/cpong and return early without updating worker status — effectively letting an external STATUS command act as a PING.
> 
> The comment doesn't covert this either
> 
> ```
> /*
>  * Process the STATUS command
>  * Load -1 : Broken
>  * Load 0  : Standby.
>  * Load 1-100 : Load factor.
>  */
> static char *process_status(request_rec *r, const char *const *ptr, int *errtype)
> {
> ```

Now I get what you mean.

This PR is only for preventing the bad load value to be set, nothing else. The correct handling of such a MCMP message is to return an error which I address in the MCMP rewrite. (I wanted this mainly because of the backport tbh.)

jajik:
Since we're not planning to backport this, I'm closing this in favor of #385

Issue #380, opened 2026-03-31
Rework MCMP handling, clarify edge cases and currently undefined behaviours


jajik:
It's already in progress, but we're not tracking it anywhere. Given this will require changes in the docs too, I'm creating this to have something we can refer to.

PR #385, opened 2026-05-15
ModCluster Management Protocol (MCMP) reimplementation

jajik:
ref #380

This PR rewrites parts of the MCMP and changes its behaviour in several cases and adds a set of tests to make sure the behaviour is correct.

Highlights:

* there's a big change in how mod_proxy_cluster's virtual hosts and aliases are handled
    - now it's one alias per each host allowing us a better (and correct) granular control of each context of each alias
* the protocol is now less permissive (e.g. bad load value in STATUS means error response)
* it's more deterministic especially when an error occurs, the state after an error are now defined
    - typically in a case of a failure during CONFIG processing we could get some parts processed and there was no indication of it
    - now there's a response indicating what failed and what did not
    - same applies to APP commands
* ping/pong routine is updated with ping now supporting http
* some refactoring and formatting took place, we're using APR API more now (e.g. when handling APR arrays), a bit more info in logging in some cases

What this PR does not include:

* final MCMP (error) messages (that's going to be addressed in #330)
* support for all output types
    - not all commands are providing all output flavours (xml), this is going to be addressed in a future PR; however, some changes here make the future implementation a bit easier
* refactoring of all APR data structures handling (again, I'll do that in a separate PR later, this PR is huge enough)
* mod_cluster_manager webpage changes to reflect changes above

I'll share a doc with more information in our team channel.

---- 
~~Side note: It seems this also resolves the instability we were seeing with the MODCLUSTER-736 test (or it's just the weather).~~ It was the weather :slightly_smiling_face: 

Side note 2: I ran the perf suite with the changes as a smoke test, it looks ok ([2 tomcats with 1 disable](https://github.com/jajik/mod_proxy_cluster-perf-comparison/actions/runs/25861578368), [20 tomcats with 10 disable](https://github.com/jajik/mod_proxy_cluster-perf-comparison/actions/runs/25862214470), [20 tomcats with 10 shutdown](https://github.com/jajik/mod_proxy_cluster-perf-comparison/actions/runs/25865374914), [20 tomcats with 10 kill](https://github.com/jajik/mod_proxy_cluster-perf-comparison/actions/runs/25865447414), [20 tomcats with 10 stop](https://github.com/jajik/mod_proxy_cluster-perf-comparison/actions/runs/25865557289))

### DIFF
--- a/native/common/common.c
+++ b/native/common/common.c
@@ -432,6 +432,7 @@ node_context *find_node_context_host(request_rec *r, const proxy_balancer *balan
             }
         }
     }
+
     if (max == 0) {
         return NULL;
     }
@@ -470,6 +471,7 @@ node_context *find_node_context_host(request_rec *r, const proxy_balancer *balan
             }
         }
     }
+
     if (nbest == 0) {
         return NULL;
     }
--- a/native/include/mod_proxy_cluster.h
+++ b/native/include/mod_proxy_cluster.h
@@ -21,7 +21,7 @@
 #include "context.h"
 #include "host.h"
 
-#define MOD_CLUSTER_EXPOSED_VERSION "mod_cluster/2.0.0.Alpha1-SNAPSHOT"
+#define MOD_CLUSTER_EXPOSED_VERSION "mod_proxy_cluster/2.0.0.Dev"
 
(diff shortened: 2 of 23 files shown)


jfclere (approved):
After a "quick" review I approve

jajik:
> After a "quick" review I approve

Thank you for the review!

PR #389, opened 2026-05-18
ci: Drop abandoned ilammy/msvc-dev-cmd in Windows job

### LINKED ISSUE #379: [CI]: Windows tooling setup deprecation
The `ilammy/msvc-dev-cmd@v1`action we're using for Windows seems to be abandoned (so it will stop working when GH moves to Node 24 in June). Find an alternative or choose a different approach for setting up the Windows build environment.

jajik:
close #379 

The `amd64` arch for Windows on ARM is intended. The Apache Lounge bits are amd64, so we're cross-compiling. It would be nice to be able to run `VsDevShell` with `-Arch amd64 -HostArch arm64`, but MS does not support that at the moment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -223,21 +223,25 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-            os: [ windows-latest, windows-11-arm ]
+            include:
+                - os: windows-latest
+                  arch: amd64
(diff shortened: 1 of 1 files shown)


rhusar (approved):

rhusar:
If it works it works, thanks @jajik

PR #394, opened 2026-08-11
Use Fedora 44 instead of 42, use `PROXY_STRNCPY` instead of `strncpy`

jajik:
The `PROXY_STRNCPY` addresses a new issue that arose with the new Fedora and `-Werror=stringop-truncation`.


rhusar (approved):
LGTM

jajik:
Thanks for the review!

PR #397, opened 2026-08-31
[Test] Fix DEBUG variable handling when defined as no/off/0

### LINKED ISSUE #396: [Test] Improper `DEBUG` definition handling
We control debugging with the `DEBUG` variable. Currently, the switch depends on the variable being defined.

That presents several issues, mainly that `DEBUG=0` or `DEBUG=off` means the debugging is ON. That also means, that our manual trigger always runs the CI in debug mode.

Ideally, we should treat `DEBUG=0`, `DEBUG=off`, and `DEBUG=Off` the same way we treat it when undefined.

jajik:
close #396 

depends on #395 (the CI is expected to fail atm)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,7 +336,7 @@ jobs:
           sh testsuite.sh
       - name: Preserve test logs
         uses: actions/upload-artifact@v6
-        if: always()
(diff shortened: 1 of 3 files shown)


jfclere (approved):
LGTM

PR #399, opened 2026-09-15
Add support for enabling/disabling mod_proxy_cluster/mod_proxy_balancer tests, split them in two in CI

### LINKED ISSUE #398: [CI] Consider running mod_proxy_cluster and mod_proxy_balancer tests in parallel
The current test suite runs for around 1.5 hours. Given we want to increase the test coverage, we should entertain the possibility of running the two parts (with mod_proxy_cluster and with mod_lbmethod_cluster/mod_proxy_balancer) in parallel.

The seemingly easiest approach is to put these two parts in an `if` and skip if some `SKIP_` variable is defined. Also, could we share the setup phase between different sections? It would save some time and resources.

jajik:
close #398

- adds new variables to control mod_proxy_cluster/mod_proxy_balancer test execution
- simplifies the default initialization of variables
- separates mod_proxy_cluster and mod_proxy_balancer tests in CI so they run in parallel (saving around 20 minutes)

We have quite a lot of jobs and artifacts at the moment. I am going to rework it a little bit so that we waste fewer resources, but I'll do that in a separate PR to keep this easy to review.


jajik:
@rhusar Unfortunately, this breaks the Required check again. Given the endless issues with this, I would suggest removing the `Required` tag for now, and I'll implement a different solution as part of the CI rework (I think the best way is to add a new `Required` job that will depend on _all_ the others – if any job fails, it will fail too).

rhusar:
> @rhusar Unfortunately, this breaks the Required check again. Given the endless issues with this, I would suggest removing the `Required` tag for now, and I'll implement a different solution as part of the CI rework (I think the best way is to add a new `Required` job that will depend on _all_ the others – if any job fails, it will fail too).

Agreed, it's annoying. The only reason it needed to be setup because other checks cannot be enforced without it...

